PR #106, opened 2026-07-11
fix(ui): ignore stale /auth/me after concurrent session change

### LINKED ISSUE #100: fix(ui): mount-time /auth/me validation can overwrite a session established by a concurrent login
## Problem

On mount, `AuthProvider` (`apps/ui/lib/auth-context.tsx:63-101`) captures `stored` from `localStorage` once, then kicks off an `/auth/me` fetch. Its `.then` handler unconditionally calls `setUser(data)` (and `setToken(stored)`) on any non-401 response, using the `stored` value captured in the closure at mount time — it does not check whether `login()`/`setSession()` have since updated `token`/`user` to a different session.

## Failure scenario

On the login page (no stored token yet, but a valid httpOnly GitHub OAuth cookie is present from a previous session), the mount-time `/auth/me` fetch is in flight. Before it resolves, the user submits the password-login form for a different account; `login()` immediately sets `token`/`user` for that account. If the slower `/auth/me` response then resolves successfully (cookie-authenticated as the *other* GitHub account), its `.then` handler calls `setUser(data)`, silently replacing the just-logged-in user with the cookie session's user, without ever touching `token`, leaving `token`/`user` inconsistent.

## Suggested fix

Guard the `/auth/me` `.then` handler with a check (e.g. an `isCancelled`/generation flag, or comparing against current `token` state) so it no-ops if a newer session was established while it was in flight.

## Severity

Low — narrow timing window requiring a stale OAuth cookie plus a concurrent password login, but produces confusing/incorrect identity state.

Tarekchehahde:
## Summary
Fixes #100.

- Track a session epoch in `AuthProvider`; bump it on `login`, `setSession`, and `logout`.
- Mount-time `/auth/me` validation no-ops when a newer session was established while the fetch was in flight.

## Test plan
Local (macOS, Node 22 + vitest):
```
vitest run tests/lib/auth-context.test.tsx
# 2 passed
tsc --noEmit
# passed
```

Covers:
1. Password login completing before a stale cookie `/auth/me` response — user stays on password account.
2. Logout before `/auth/me` resolves — user stays logged out.

GitHub CI runs UI typecheck, lint, vitest, and build on PR.

### DIFF
--- a/apps/ui/lib/auth-context.tsx
+++ b/apps/ui/lib/auth-context.tsx
@@ -1,6 +1,6 @@
 "use client"
 
-import { createContext, useCallback, useContext, useEffect, useState } from "react"
+import { createContext, useCallback, useContext, useEffect, useRef, useState } from "react"
 
 export interface AuthUser {
   id: number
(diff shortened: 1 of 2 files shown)


nazarli-shabnam (approved):

nazarli-shabnam:
@Tarekchehahde Well done!

PR #112, opened 2026-07-11
fix(api): reject Organization account_type on personal installation sync (#102)

### LINKED ISSUE #102: fix(installations): enforce ownership check regardless of account_type on /me/installations/sync
## Problem

`sync_personal_installation` (`apps/api/src/routers/installations.py:65-91`) only validates the caller's identity when `payload.account_type == "User"` (checking `account_login` matches the caller's own GitHub login). `SyncInstallationsInput.account_type` (`apps/api/src/schemas/installation.py:6-10`) defaults to `"Organization"`, so a caller can simply omit `account_type` (or pass anything other than `"User"`) to skip validation entirely and register a `GitHubInstallation` row with an arbitrary `account_login` and `installation_id` under their own `owner_user_id`.

## Failure scenario

Any authenticated user POSTs `/me/installations/sync` with `{"account_login": "some-other-org", "installation_id": <victim's real installation id>}` (default `account_type` applies). The endpoint happily creates a personal installation row pointing at an org/installation the caller has no relationship to, with no server-side verification against GitHub. This is exactly the check that exists for the `"User"` branch but is silently skipped for every other value.

## Suggested fix

Require verification (or reject) for any `account_type` on `/me/installations/sync`, not just `"User"` — e.g. only allow `account_type == "User"` on this endpoint and reject `"Organization"` outright (org installs should go through the org-scoped `/orgs/{org_login}/installations/sync` path, which does call `assert_owner_matches_org`).

## Severity

Medium-High — currently limited blast radius since `get_installation_token()` isn't yet wired to these rows, but it lets any user record arbitrary org/installation associations under their account, and would become a full privilege-escalation path the moment installation tokens are minted from these rows.

Tarekchehahde:
## Summary
Fixes #102.

Personal installation sync now rejects `account_type: Organization` with a 403 instead of silently accepting it.

## Test plan
- [x] `pytest apps/api/tests/test_installations.py::test_sync_personal_installation_rejects_organization_account_type -v`

### DIFF
--- a/apps/api/src/routers/installations.py
+++ b/apps/api/src/routers/installations.py
@@ -68,18 +68,22 @@ def sync_personal_installation(
     user: UserOut = Depends(require_auth),
     db: Session = Depends(get_db),
 ):
-    if payload.account_type == "User":
-        db_user = db.query(User).filter(User.id == user.id).first()
-        if not db_user or not db_user.github_login:
-            raise HTTPException(
-                status_code=status.HTTP_403_FORBIDDEN,
-                detail="Link your GitHub account before syncing a personal installation",
-            )
-        if payload.account_login != db_user.github_login:
-            raise HTTPException(
-                status_code=status.HTTP_403_FORBIDDEN,
-                detail="account_login must match your own GitHub account",
-            )
+    if payload.account_type != "User":
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Personal installation sync only supports account_type User; use org sync for organizations",
+        )
+    db_user = db.query(User).filter(User.id == user.id).first()
+    if not db_user or not db_user.github_login:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Link your GitHub account before syncing a personal installation",
+        )
+    if payload.account_login != db_user.github_login:
+        raise HTTPException(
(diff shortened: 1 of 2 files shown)


nazarli-shabnam (changes requested):
Hey @Tarekchehahde — two things here.

**1. Test doesn't match the code.** This PR adds `test_sync_org_installation_upserts_existing_row`, expecting org sync to upsert an existing row instead of inserting a duplicate. I ran it against this branch on its own and it fails: two rows get created instead of one being updated (`assert 2 == 1`). That upsert logic actually lives in #120, not here — looks like this test got picked up from that PR by mistake. Can you drop it from this PR? The account_type→403 rejection itself is correct and well-tested by the other test.

**2. PR description got mangled.** The body currently renders as one big line with literal `\n` text in it instead of real line breaks (and a couple of missing words — "Personal  now rejects  with 403"). Because of that, `Fixes #102` probably won't get picked up as a closing keyword by GitHub either. Here's a clean version you can paste in:

```
## Summary
Fixes #102.

Personal installation sync now rejects `account_type: Organization` with a 403 instead of silently accepting it.

## Test plan
- `pytest apps/api/tests/test_installations.py::test_sync_personal_installation_rejects_organization_account_type -v`
```

Tarekchehahde:

nazarli-shabnam (approved):

PR #113, opened 2026-07-11
fix(auth): return 401 for password login on GitHub-only accounts (#101)

### LINKED ISSUE #101: fix(auth): handle password login for GitHub-only accounts instead of crashing
## Problem

`_verify_password(password, password_hash)` (`apps/api/src/routers/auth.py:36-41`) calls `password_hash.encode()`, but `User.password_hash` is nullable and is `None` for any account created via GitHub OAuth (`apps/api/src/routers/github_auth.py:53-61` sets `password_hash=None`). In `login()` (lines 153-160), `if not user or not _verify_password(...)` calls `_verify_password` for any existing user, including GitHub-only ones, before checking that a password hash exists.

## Failure scenario

A user who signed up via "Sign in with GitHub" (no password set) later tries POST /auth/login with their email and any password. `bcrypt.checkpw(password.encode(), None.encode())` raises `AttributeError`, producing an unhandled 500 instead of the documented "Always returns 401 on failure (no field leaking)" behavior — and the 500-vs-401 status difference itself leaks whether an email belongs to a GitHub-only account (user enumeration), contradicting the docstring's stated guarantee.

## Suggested fix

In `login()`, short-circuit to 401 when `user.password_hash is None` before calling `_verify_password`, so the response is indistinguishable from "wrong password".

## Severity

High — unauthenticated endpoint crashes (reliability) and leaks account existence/type (security), contradicting the code's own documented invariant.

Tarekchehahde:
## Summary
Fixes #101.

Password login on GitHub-only accounts now returns 401 with a clear message instead of a generic 500.

## Test plan
- [x] `pytest apps/api/tests/test_auth.py -v -k github_only`

### DIFF
--- a/apps/api/src/routers/auth.py
+++ b/apps/api/src/routers/auth.py
@@ -37,7 +37,9 @@ def _hash_password(password: str) -> str:
     return bcrypt.hashpw(password.encode(), bcrypt.gensalt()).decode()
 
 
-def _verify_password(password: str, password_hash: str) -> bool:
+def _verify_password(password: str, password_hash: str | None) -> bool:
+    if not password_hash:
+        return False
     return bcrypt.checkpw(password.encode(), password_hash.encode())
 
 
--- a/apps/api/tests/test_auth.py
+++ b/apps/api/tests/test_auth.py
@@ -142,6 +142,18 @@ def test_login_unknown_email(auth_client):
     assert resp.status_code == 401
 
 
+def test_login_github_only_user_returns_401(auth_client, db):
+    from src.core.db import User
+
+    db.add(User(email="github-only@example.com", password_hash=None, is_workspace_admin=False))
+    db.commit()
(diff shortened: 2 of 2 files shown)


nazarli-shabnam (changes requested):
Hey @Tarekchehahde — code and tests here are solid, ran clean in isolation, no issues there.

Just flagging that the PR description rendered as one long line with literal `\n` characters instead of actual line breaks, which likely stops GitHub from picking up `Fixes #101` as a closing keyword. Here's a clean version to paste in:

```
## Summary
Fixes #101 (same root cause as #76).

GitHub-OAuth-only accounts no longer 500 on password login — API returns 401 instead.

## Test plan
- `pytest apps/api/tests/test_auth.py -v -k github_only`
```

Tarekchehahde:

nazarli-shabnam (approved):

Issue #96, opened 2026-07-11
fix(ui): del() HTTP helper skips 401 handling that all other verbs perform


nazarli-shabnam:
## Problem

`get`, `post`, `put`, and `patch` in `apps/ui/lib/api/client.ts` all funnel through `handleResponse`, which on a 401 clears the stale token from localStorage and dispatches `clevis:unauthorized` so `AuthGuard` redirects to `/login`. `del()` (lines 98-107) reimplements its own response handling and never checks `res.status === 401`, so it never clears the token or fires the event.

## Failure scenario

A user's JWT expires while on Settings, and they click delete on a saved token row (`api.tokens.delete`) or otherwise trigger a DELETE call. The request 401s, the caller only sees a generic "Request failed: 401" error, the stale token remains in localStorage, and the app never redirects to `/login` — leaving the user stuck with dead auth state until they refresh or navigate manually.

## Suggested fix

Route `del()` through the same `handleResponse`-style 401 handling (or extract a shared `checkAuth(res)` helper used by all five verbs).

## Severity

Medium — breaks the expired-session recovery UX and leaves a stale token lingering in localStorage.

Tarekchehahde:
Addressed in batch PR #109 (local tests in commit message).

Tarekchehahde:
PR: https://github.com/nazarli-shabnam/clevis/pull/114

nazarli-shabnam:
Already fixed in the current code — verified by reading the exact lines this issue cites.

\`apps/ui/lib/api/client.ts\`'s \`del()\` already checks \`res.status === 401\` and mirrors \`handleResponse\`'s handling:
\`\`\`ts
if (res.status === 401) {
  if (typeof window !== "undefined") localStorage.removeItem(_TOKEN_KEY)
  window.dispatchEvent(new Event("clevis:unauthorized"))
}
\`\`\`
Closing.

PR #114, opened 2026-07-11
fix(ui): handle 401 in del() HTTP helper like other verbs (#96)

Tarekchehahde:
## Summary\nFixes #96.\n\n`del()` in the API client now clears auth state on 401, matching get/post/patch/put.\n\n## Test plan\n- [x] `cd apps/ui && bun run typecheck`

### DIFF
--- a/apps/ui/lib/api/client.ts
+++ b/apps/ui/lib/api/client.ts
@@ -100,6 +100,10 @@ async function del(path: string): Promise<void> {
     method: "DELETE",
     headers: { ...getAuthHeaders() },
   })
+  if (res.status === 401) {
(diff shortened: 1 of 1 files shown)


nazarli-shabnam (changes requested):
Hey @Tarekchehahde — the 401 handling in `del()` is correct and exactly what's needed here.

But this diff also carries the `encodeURIComponent()` changes to `cache.list`/`cache.clear` — that's #122's fix, not this one, and it'll conflict with #122 when both land. Mind rebasing those bits out so this PR stays scoped to the `del()` fix? Looks like it's just the second change block in `client.ts`.

Tarekchehahde:

nazarli-shabnam (approved):

PR #116, opened 2026-07-11
fix(checks): isolate failures so one check cannot abort the run (#81)

### LINKED ISSUE #81: fix(checks): a single failing check aborts the entire run, no per-check isolation
## Problem

`run_all_checks()` (`packages/checks/src/checks/runner.py:10-19`) loops over the three `Check` subclasses and calls `check.run(...)` with no per-check try/except. Any single check raising — a GitHub rate limit (403/429 via `raise_for_status()`), the `security_and_analysis: null` bug (see related issue), or `OrgMFARequired` needing org-owner scope that a lesser-privileged token doesn't have — aborts the loop and fails the whole request, even when the other checks would have succeeded independently.

## Failure scenario

An org has 50 repos; check 1 and 2 succeed, check 3 hits a transient GitHub rate limit on repo 40. The entire `run_all_checks` call throws, and `analytics_service.py` (which depends on this returning complete results to compute the score) gets nothing — the user sees a full failure instead of a partial score with one check flagged as unavailable.

## Suggested fix

Wrap each `check.run(...)` call in its own try/except inside the loop, and represent a failed check as a distinct result status (e.g. `"status": "error"`) rather than crashing the whole batch, so partial results and a partial score are still returned.

## Severity

Medium — turns isolated, expected failure modes (rate limits, scope gaps) into total outages for analytics.

Tarekchehahde:
## Summary
Fixes #81.

Wrap each check in the runner so one failing check cannot abort the entire run.

## Test plan
- [x] `pytest packages/checks/tests/test_runner.py -v`

### DIFF
--- a/apps/api/src/services/analytics_service.py
+++ b/apps/api/src/services/analytics_service.py
@@ -7,7 +7,7 @@ def get_overview(owner: str, token: str) -> dict:
     base_url = settings.github_api_base
     report = run_all_checks(owner=owner, token=token, base_url=base_url)
     checks = report["checks"]
-    failed = [c for c in checks if c["status"] == "fail"]
+    failed = [c for c in checks if c["status"] in ("fail", "error")]
     score = 100 - int((len(failed) / max(1, len(checks))) * 100)
     return {
         "owner": owner,
--- a/apps/api/tests/test_analytics_scoring.py
+++ b/apps/api/tests/test_analytics_scoring.py
@@ -17,4 +17,4 @@ def test_overview_counts_error_checks_against_score():
         overview = get_overview(owner="acme", token="tok")
 
     assert overview["failed_checks"] == 2
-    assert overview["score"] == 33
+    assert overview["score"] == 34
--- a/apps/api/tests/test_analytics_scoring.py
+++ b/apps/api/tests/test_analytics_scoring.py
@@ -1,0 +1,20 @@
+"""Analytics score treats errored checks as failures, not silent passes."""
+
+from unittest.mock import patch
+
+from src.services.analytics_service import get_overview
+
+
+def test_overview_counts_error_checks_against_score():
+    errored = {
+        "checks": [
+            {"status": "pass"},
+            {"status": "error"},
+            {"status": "fail"},
+        ]
+    }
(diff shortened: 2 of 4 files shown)


nazarli-shabnam (changes requested):
Hey @Tarekchehahde — logic and tests here checked out clean in isolation, no problems with the fix itself.

Just the PR description: it's rendering as one long line with literal `\n` in it instead of real line breaks, which likely blocks GitHub from recognizing `Fixes #81` as a closing keyword. Clean version to paste in:

```
## Summary
Fixes #81.

`run_all_checks` wraps each check in try/except and reports errors per-check instead of aborting the whole run.

## Test plan
- `pytest packages/checks/tests/test_runner_isolation.py -v`
```

Tarekchehahde:

nazarli-shabnam (changes requested):
The narrow fix is correct: run_all_checks() now wraps each check.run(...) in try/except, logs the exception, and reports {"status": "error", "value": "Check failed: <id>"} instead of letting the exception propagate and abort the whole batch. Confirmed against current main — the file matches exactly what issue #81 describes (no try/except at all), and the new test (test_run_all_checks_continues_when_one_check_raises) genuinely exercises the isolation: one check raises, others still return results, and the errored one gets status: "error".

But it doesn't fully honor what the issue actually asked for. I traced the new "error" status downstream into apps/api/src/services/analytics_service.py:10:
failed = [c for c in checks if c["status"] == "fail"]
score = 100 - int((len(failed) / max(1, len(checks))) * 100)
This only filters on status == "fail". A check that errors (rate limit, missing scope, etc.) is counted in the denominator (len(checks)) but not in failed — so it's silently treated as if it. That directlycontradicts the issue's own stated goal: "the user sees a full failure instead check flagged asunavailable" — the intent was for an errored check to be visibly reflecte, not counted as a clean pass.

So: the PR stops the crash (the literal bug), but introduces a silent scoring bug as a side effect — a rate-limited check now quietly boosts the score instead of "flagging as unavailable." analytics_service.py wasn't touched by this PR at all, so this gap was never addressed.

Tarekchehahde:

nazarli-shabnam (approved):

PR #117, opened 2026-07-11
chore(api,worker): dedupe _crypto.py into shared checks.crypto module (#83)

### LINKED ISSUE #83: chore(api,worker): _crypto.py is duplicated between apps/api and apps/worker with no shared source
## Problem

`apps/api/src/core/_crypto.py` and `apps/worker/src/_crypto.py` are byte-for-byte identical implementations of Fernet key derivation and encrypt/decrypt, maintained as two independent copies rather than a shared package (unlike `packages/checks`, which both services could depend on).

## Failure scenario

A future change to one copy — adding key rotation, switching the KDF, adding a salt (see the related key-derivation issue) — that isn't mirrored exactly in the other copy silently breaks encrypt/decrypt compatibility: tokens encrypted by the API (in `jobs.payload` or `saved_tokens`) become undecryptable by the worker, with no compile-time or CI signal that the two files have drifted.

## Suggested fix

Extract `_crypto.py` into a small shared package (similar to `packages/checks`) that both `apps/api` and `apps/worker` install as a dependency, so there is a single source of truth.

## Severity

Medium — no active bug today, but a correctness/reliability trap for the next change in this area.

Tarekchehahde:
## Summary
Fixes #83.

Deduplicate `_crypto.py` into a shared `checks.crypto` module used by both API and worker.

## Test plan
- [x] `pytest apps/api/tests/test_crypto.py apps/worker/tests/ -v`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -32,7 +32,7 @@ pnpm-debug.log*
 .agents/
 .github/
 
-# Image that COPYs only apps/api + packages/checks
+# Images that COPY only apps/api (worker Dockerfile also uses repo root context)
 apps/ui/
 deploy/
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Build + Push Worker
         uses: docker/build-push-action@v6
         with:
-          context: apps/worker
+          context: .
           file: apps/worker/Dockerfile
           push: true
           tags: |
--- a/apps/api/src/core/_crypto.py
+++ b/apps/api/src/core/_crypto.py
@@ -1,37 +1,3 @@
-import base64
-import hashlib
+from checks.crypto import decrypt_job_token, encrypt_job_token
 
-from cryptography.fernet import Fernet
-from cryptography.hazmat.primitives import hashes
-from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
-
-# Fixed application-level salt for the PBKDF2 key derivation below. It isn't secret — its
-# job is to make the derived key resistant to precomputed-hash attacks, not to add
-# per-ciphertext randomness (JOB_SECRET_KEY itself is what must stay secret).
-_PBKDF2_SALT = b"clevis-job-token-fernet-v2"
-_PBKDF2_ITERATIONS = 480_000
-_V2_PREFIX = "v2:"
-
-
-def _legacy_fernet(raw_key: str) -> Fernet:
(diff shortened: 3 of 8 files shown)


nazarli-shabnam (changes requested):
Hey @Tarekchehahde — re-checked this one in isolation, dedupe is clean, nothing wrong with the code.

Same formatting issue as a couple others though: the PR body has literal `\n` characters instead of real line breaks, which will likely stop GitHub from recognizing `Fixes #83` as a closing keyword. Clean version to paste in:

```
## Summary
Fixes #83.

Extracts the shared Fernet helpers into `packages/checks/src/checks/crypto.py`; the API and worker now re-export from there instead of maintaining duplicate copies.

## Test plan
- `pytest packages/checks/tests -v`
```

Tarekchehahde:

nazarli-shabnam (changes requested):
The dedupe logic itself is correct and matches issue #83: both _crypto.py files now just re-export decrypt_job_token/encrypt_job_token from the new packages/checks/src/checks/crypto.py, which is byte-for-byte the same logic that was duplicated before (confirmed against current main, which still has the old duplicated implementation in both places).

But the worker's Dockerfile was never updated to install the new dependency, and this will break the worker in production.

- apps/api/Dockerfile (lines 9-10) does COPY packages/checks /app/packages/checks + pip install /app/packages/checks — the API side is fine.
- apps/worker/Dockerfile does not copy or install packages/checks at all — it only runs pip install -r requirements.txt, and neither apps/worker/requirements.txt nor the Dockerfile references the checks package.
- The PR's new apps/worker/src/_crypto.py does from checks.crypto import decrypt_job_token, encrypt_job_token. In the actual worker Docker image, that import will raise ModuleNotFoundError: No module named 'checks' — the worker process would crash as soon as it tries to decrypt a job token (i.e., on its very first real job).

Why CI is green anyway: the "Docker Build Verification" job only runs docker build — it never runs the built image, so a missing runtime import never surfaces there. The "Python Tests + Compile" job does pip install -e packages/checks directly in the CI environment (not via the worker's Dockerfile) before running tests/compileall, which is why tests and compile both pass despite the Docker image being broken. Local dev also masks this, since the one-time setup instructions in CLAUDE.md have you pip install -e packages/checks globally.

This is a real deploy-breaking gap, not a nitpick — worth flagging as a blocking comment (add COPY packages/checks + pip install to apps/worker/Dockerfile, mirroring the API's Dockerfile) before merge.

Tarekchehahde:

nazarli-shabnam (approved):

PR #118, opened 2026-07-11
fix(ui): per-row pending state for member revoke mutation (#88)

### LINKED ISSUE #88: fix(ui): shared revoke mutation disables every member row at once
## Problem

`apps/ui/app/settings/org/[login]/members/page.tsx:34-37,105-107` backs every row's "Revoke" button with a single shared `useMutation` instance. `disabled={revoke.isPending}` therefore disables *all* rows while any one revoke is in flight, with no per-row indicator of which specific invitation is being revoked.

## Failure scenario

An admin managing a long list of pending invitations clicks "Revoke" on one row; every other row's button also greys out until that single request completes, and there's no visual cue (spinner, etc.) tying the pending state to the specific row that was clicked — confusing on a slow connection or a long list.

## Suggested fix

Key the mutation (or its pending state) per invitation id, e.g. via `mutate(id)` combined with tracking the in-flight id in local state, so only the row being revoked shows a pending/disabled state.

## Severity

Low — UX confusion only, not a correctness or security issue.

Tarekchehahde:
## Summary
Fixes #88.

Track pending state per row for member revoke mutations so only the clicked row shows a spinner.

## Test plan
- [x] `cd apps/ui && bun run test`

### DIFF
--- a/apps/ui/app/settings/org/[login]/members/page.tsx
+++ b/apps/ui/app/settings/org/[login]/members/page.tsx
@@ -22,6 +22,8 @@ export default function OrgMembersPage() {
     queryFn: () => api.invitations.list(orgLogin),
   })
 
+  const [revokingId, setRevokingId] = useState<number | null>(null)
+
   const invite = useMutation({
     mutationFn: () => api.invitations.create(orgLogin, email.trim()),
     onSuccess: (data) => {
@@ -33,6 +35,8 @@ export default function OrgMembersPage() {
 
   const revoke = useMutation({
     mutationFn: (id: number) => api.invitations.revoke(orgLogin, id),
+    onMutate: (id) => setRevokingId(id),
+    onSettled: () => setRevokingId(null),
     onSuccess: () => queryClient.invalidateQueries({ queryKey: ["invitations", orgLogin] }),
   })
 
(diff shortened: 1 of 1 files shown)


nazarli-shabnam (changes requested):
Hey @Tarekchehahde — re-tested this in isolation, the per-row pending state fix is correct, no code issues.

Only the description needs a fix: it's rendering as one long line with literal `\n` in it, which likely stops GitHub from recognizing `Fixes #88` as a closing keyword. Clean version to paste in:

```
## Summary
Fixes #88.

Revoke button's pending state is now scoped to the row being revoked, not all rows.

## Test plan
- `cd apps/ui && bun run typecheck`
```

Tarekchehahde:

nazarli-shabnam (approved):
The fix correctly resolves issue #88 (low severity, UX-only): confirmed on main that revoke was a single shared useMutation, and disabled={revoke.isPending} on every row meant clicking revoke on one invitation greyed out every row's button with no indication of which one was actually being revoked.

The PR adds a revokingId state, set in onMutate and cleared in onSettled, and switches the disabled check to revokingId === inv.id. This scopes the pending state to the clicked row exactly as the issue's suggested fix describes — other rows stay interactive while one revoke is in flight.

PR #119, opened 2026-07-11
fix(ui): surface logout network failure instead of swallowing it (#89)

### LINKED ISSUE #89: fix(ui): logout swallows network failure, leaving server-side session alive
## Problem

`logout()` (`apps/ui/lib/auth-context.tsx:52-58`) fires `fetch(.../auth/logout, { credentials: "include" }).catch(() => {})` and then unconditionally clears local state (`localStorage`, `token`, `user`) regardless of whether the server-side call succeeded.

## Failure scenario

A user on a flaky connection clicks "Log out." The `/auth/logout` request fails to reach the server (network blip), but local state is cleared anyway and the UI shows the user as logged out. For GitHub-OAuth sessions (httpOnly cookie), the server-side session cookie was never actually invalidated — the user believes they've logged out, but the cookie remains live, e.g. on a shared/public machine.

## Suggested fix

Surface a non-blocking warning if the logout network call fails (local state can still be cleared for UX purposes, but the user should be told the server-side session may still be active), or retry the logout call before finishing.

## Severity

Low — narrow window (network failure at the exact moment of logout) but has real security implications on shared devices.

Tarekchehahde:
## Summary
Fixes #89.

When `/auth/logout` fails over the network, show a warning banner on the login page so users know the server session may still be active.

## Test plan
- [x] `cd apps/ui && bun run typecheck && bun run test`
- [x] Manual: trigger logout with API unreachable → warning appears on `/login`

### DIFF
--- a/apps/ui/app/login/page.tsx
+++ b/apps/ui/app/login/page.tsx
@@ -6,7 +6,7 @@ import { useAuth } from "@/lib/auth-context"
 import { api } from "@/lib/api/client"
 import { Input } from "@/components/ui/input"
 import { Button } from "@/components/ui/button"
-import { Loader2 } from "lucide-react"
+import { Loader2, AlertTriangle } from "lucide-react"
 
 const API_BASE = process.env.NEXT_PUBLIC_API_BASE || "http://localhost:8080"
 
@@ -40,7 +40,7 @@ function githubErrorMessage(code: string | null): string {
 }
 
 export default function LoginPage() {
-  const { user, login, isLoading } = useAuth()
+  const { user, login, isLoading, logoutWarning, clearLogoutWarning } = useAuth()
   const router = useRouter()
   const searchParams = useSearchParams()
   const next = safeNextPath(searchParams.get("next"))
@@ -79,6 +79,7 @@ export default function LoginPage() {
   async function handleSubmit(e: React.FormEvent) {
     e.preventDefault()
     setError("")
+    clearLogoutWarning()
     setIsSubmitting(true)
     try {
       await login(email, password)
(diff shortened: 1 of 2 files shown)


nazarli-shabnam (changes requested):
Hey @Tarekchehahde — the state you're adding (`logoutWarning`, `clearLogoutWarning`) looks right, but this PR only touches `auth-context.tsx` — nothing in the UI actually reads `logoutWarning` anywhere, so right now the warning gets computed and then goes nowhere.

Can you add the piece that actually surfaces it (toast, banner, whatever we use elsewhere) so this really does what the title says? The context plumbing itself is good, just needs the consumer.

Tarekchehahde:

nazarli-shabnam (approved):
The fix genuinely closes issue #89, and addresses the reviewer's exact follow-up request:
- Issue #89: logout() fires /auth/logout and swallows any failure (.catch(() => {})), then unconditionally clears local state — so on a network blip during logout, the server-side session (particularly the GitHub OAuth httpOnly cookie) stays alive while the UI shows the user as logged out, with no warning. Confirmed this exact behavior on current main.
- First commit added the state plumbing (logoutWarning, clearLogoutWarning) to auth-context.tsx — logout() now checks res.ok and sets a warning message on both HTTP failure and network .catch.
- Your review correctly caught that nothing consumed logoutWarning — it was computed and went nowhere.
- The second commit (fix(ui): show logout network warning on login page) is the actual fix for your comment: login/page.tsx now destructures logoutWarning/clearLogoutWarning, renders a banner when it's set, and clears it on the next login attempt (clearLogoutWarning() in handleSubmit). I checked this against the real diff — it's there, not just claimed.
- Since AuthProvider wraps the whole app, logoutWarning persists in memory across the client-side redirect from wherever logout() was triggered to /login, so this works for the general flow (not just a literal logout button on the login page itself).

PR #120, opened 2026-07-11
fix(api): upsert installation sync rows instead of always inserting (#79)

### LINKED ISSUE #79: fix(api): installation /sync always inserts, never upserts — creates duplicate rows
## Problem

`installation_repo.create()` (`apps/api/src/repositories/installation_repo.py:6-30`) unconditionally does `db.add(row); db.commit()` with no lookup/update by `(org_id, account_login)` or `(owner_user_id, account_login)` first, and there is no unique constraint on those columns in `core/db.py`. The `/orgs/{org}/installations/sync` and `/me/installations/sync` routes are documented (per the routers' own docstrings) as re-syncing metadata for an *existing* installation, but every call inserts a brand-new `github_installations` row instead of updating the existing one.

## Failure scenario

A workspace admin (or automated process) calls `/sync` more than once for the same org/account — e.g. to refresh `token_ref` after rotating a token. Each call adds a new row. `list_for_org`/`list_for_user` now return duplicates for the same account, and it becomes ambiguous which row's `token_ref` is authoritative — different code paths could read different (stale vs current) rows for the same logical installation.

## Suggested fix

Add a unique constraint on `(org_id, account_login)` / `(owner_user_id, account_login)` and change `create()` to an upsert (update-if-exists, insert-if-not) keyed on that constraint.

## Severity

Medium — data integrity issue that can cause token/installation state to silently diverge.

Tarekchehahde:
## Summary
Fixes #79.

Upsert installation sync rows instead of always inserting duplicates. Adds partial unique indexes for org/user account keys.

## Test plan
- [x] `pytest apps/api/tests/test_installations.py::test_sync_org_installation_upserts_existing_row -v`

### DIFF
--- a/apps/api/alembic/versions/0013_installation_unique_constraints.py
+++ b/apps/api/alembic/versions/0013_installation_unique_constraints.py
@@ -1,0 +1,36 @@
+"""Unique constraints for github_installations upsert keys.
+
+Revision ID: 0013
+Revises: 0012
+Create Date: 2026-07-11
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "0013"
+down_revision = "0012"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_index(
+        "uq_github_installations_org_account",
+        "github_installations",
+        ["org_id", "account_login"],
+        unique=True,
+        postgresql_where=sa.text("org_id IS NOT NULL"),
+    )
+    op.create_index(
+        "uq_github_installations_user_account",
+        "github_installations",
+        ["owner_user_id", "account_login"],
+        unique=True,
+        postgresql_where=sa.text("owner_user_id IS NOT NULL"),
+    )
+
+
+def downgrade() -> None:
+    op.drop_index("uq_github_installations_user_account", table_name="github_installations")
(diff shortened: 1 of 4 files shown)


nazarli-shabnam (changes requested):
Hey @Tarekchehahde — the upsert logic here looks right, but I ran the new test you added (`test_sync_personal_installation_rejects_organization_account_type`) against this branch on its own and it fails: `assert 200 == 403`.

That rejection (blocking `account_type: Organization` on the personal sync route) is actually implemented over in #112, not here — this PR never touches `routers/installations.py`. Since the test depends on a fix that lives in a different PR, it'll go red in CI if #120 lands before/without #112.

Can you either drop that one test from this PR (it belongs with #112) or rebase on top of #112 once that's merged? Everything else here — the upsert repo logic, the migration — looks solid.

Tarekchehahde:

nazarli-shabnam (dismissed):

nazarli-shabnam:
@Tarekchehahde fix the conflicts please. Other than that, its ready to merge.

Tarekchehahde:

nazarli-shabnam (approved):

Issue #22, opened 2026-04-23
worker refactor + queue robustness.


nazarli-shabnam:

Tarekchehahde:
Addressed in batch PR #109 (local tests in commit message).

Tarekchehahde:
PR: https://github.com/nazarli-shabnam/clevis/pull/121

Tarekchehahde:
Updated PR #121: rebased onto upstream main (includes migration 0013). Retry cap moved to separate PR for #130.

PR #121, opened 2026-07-11
fix(worker): reclaim stale processing jobs and validate payloads (#22)

Tarekchehahde:
## Summary
Partial fix for #22.

Worker reclaims jobs stuck in `processing` past a timeout, validates job payloads before execution, and caps reclaim retries via `jobs.retry_count` (migration `0014`).

## Test plan
- [x] `pytest apps/worker/tests/test_process_job.py -v` (7 passed)

### DIFF
--- a/apps/api/alembic/versions/0013_add_job_retry_count.py
+++ b/apps/api/alembic/versions/0013_add_job_retry_count.py
@@ -1,0 +1,25 @@
+"""Add jobs.retry_count for stale-processing reclaim cap.
+
+Revision ID: 0013
+Revises: 0012
+Create Date: 2026-07-12
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "0013"
+down_revision = "0012"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "jobs",
+        sa.Column("retry_count", sa.Integer(), nullable=False, server_default="0"),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("jobs", "retry_count")
--- a/apps/api/alembic/versions/0014_add_job_retry_count.py
+++ b/apps/api/alembic/versions/0014_add_job_retry_count.py
@@ -1,15 +1,15 @@
 """Add jobs.retry_count for stale-processing reclaim cap.
 
-Revision ID: 0013
-Revises: 0012
+Revision ID: 0014
+Revises: 0013
 Create Date: 2026-07-12
 """
 
 import sqlalchemy as sa
 from alembic import op
 
-revision = "0013"
-down_revision = "0012"
+revision = "0014"
+down_revision = "0013"
 branch_labels = None
 depends_on = None
 
--- a/apps/api/src/core/db.py
+++ b/apps/api/src/core/db.py
@@ -79,6 +79,7 @@ class Job(Base):
     payload: Mapped[str] = mapped_column(Text, nullable=False)
     status: Mapped[str] = mapped_column(String, nullable=False, default="queued")
     result: Mapped[str | None] = mapped_column(Text, nullable=True)
+    retry_count: Mapped[int] = mapped_column(Integer, nullable=False, server_default="0")
     created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now())
     updated_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), onupdate=func.now())
 
--- a/apps/worker/src/worker.py
+++ b/apps/worker/src/worker.py
@@ -17,6 +17,8 @@ log = logging.getLogger(__name__)
 
 _DB_URL = settings.database_url.get_secret_value().replace("postgresql+psycopg://", "postgresql://")
 _PROCESSING_STALE_MINUTES = 30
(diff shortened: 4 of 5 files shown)


nazarli-shabnam (changes requested):
Hey @Tarekchehahde — re-ran the worker tests against this branch in isolation, all pass, the stale-reclaim + payload-validation logic looks right.

Just the description: it's got literal `\n` characters instead of real line breaks. Since this is explicitly a partial fix, I kept the "Partial fix for #22" wording rather than a closing keyword — didn't want it auto-closing an issue that isn't fully done. Clean version to paste in:

```
## Summary
Partial fix for #22.

Worker reclaims jobs stuck in `processing` past a timeout and validates job payloads before execution.

## Test plan
- `pytest apps/worker/tests/test_process_job.py -v`
```

Tarekchehahde:

nazarli-shabnam (changes requested):
Two of the three problems in issue #22 are correctly fixed:
- Stuck jobs on crash — confirmed on main: no reclaim logic existed anywhere in worker.py, and the poll query only ever selects status = 'queued', so a job left in processing by a crashed worker was stuck permanently. The PR adds _reclaim_stale_jobs(), run once per poll loop before claiming the next job, resetting anything stuck in processing for >30 minutes back to queued. Correct mechanism.
- Unvalidated payloads — confirmed on main: process_job did payload["owner"], payload["repo"], decrypt_job_token(payload["token"], ...) directly with zero validation, so a malformed row surfaced as an opaque KeyError/TypeError. The PR adds _validate_payload() (checks owner/repo/token are present and non-empty) plus a dict type check, run before decryption/network I/O — exactly matching the issue's suggested ordering. The new test genuinely exercises this (process_job(conn, 6, json.dumps({"owner": "acme"})) → asserts status='failed' and the specific error message), not just a smoke test.

The third problem — retry/backoff for transient failures — isn't attempted at all, and that's fine; the reviewer explicitly signed off on this being intentionally partial.

But there's a real gap the reviewer didn't call out, and their one requested change wasn't applied:
1. Issue #22's suggested fix for the reclaim logic specifically says it should be "guarded by a retry-count column so a job that repeatedly crashes its worker doesn't loop forever (cap retries, then mark failed with a clear reason)." This PR's _reclaim_stale_jobs() has no retry-count column and no cap — a job that reliably crashes its worker (e.g., a genuine poison-pill payload that gets past validation, or an environment-level crash) would be reclaimed and reprocessed forever, never landing in failed. This is a real, unaddressed piece of the reclaim design, not just the separately-scoped retry/backoff item.
2. The reviewer's actual requested change — reword the PR body from Fixes #22 to Partial fix for #22, specifically so merging doesn't auto-close an issue that's only ⅔ done — has not been applied. I checked the current PR body directly: it still reads Fixes #22. verbatim. The formatting (literal \n) got fixed, but the substantive part of the ask (the closing keyword) was not. Merging as-is will auto-close #22 while the retry/backoff item (and the retry-cap gap above) remain outstanding.

Tarekchehahde:

nazarli-shabnam:
@Tarekchehahde see tests failing and solve conflicts in your other prs as well. Also consider to add your name to commits.

Tarekchehahde:
@nazarli-shabnam Rebased on latest `main`: retry cap + migration `0014` included, **7/7** worker tests pass on web2py 3.3.3 stack. PR body updated to **Partial fix for #22** (no closing keyword). Signed-off-by on commits. Ready for re-review.

nazarli-shabnam:
Tag me when they are ready for review (ci passing and completely fixes the mentioned issues). @Tarekchehahde

nazarli-shabnam (changes requested):
Appreciate that this is honestly labeled "Partial fix for #22" rather than claiming to close it — that's the right way to scope a multi-part issue. Two things before this can merge, though:

**1. CI is red.** Diff coverage is 89%, below the 90% floor:
```
apps/worker/src/worker.py (88.9%): Missing lines 81,123
```
Line 81 is inside `_reclaim_stale_jobs`/the new validation path and line 123 is in `run()`'s reclaim-call wiring — worth adding coverage there (or checking if `_read_app_config`'s docstring removal shifted line numbers and something else got missed).

**2. Given this is explicitly a partial fix, worth confirming what's left is tracked.** #22 lists three problems: (1) stuck jobs on crash, (2) unvalidated payloads, (3) no retry/backoff for transient failures. This PR does (1) and (2) well — the reclaim query correctly distinguishes "requeue with retry_count+1" from "cap and fail after `_MAX_RECLAIM_RETRIES`," and payload validation happens before decrypt/network I/O as the issue asked. But (3) isn't touched at all: `process_job`'s except block (worker.py:73-79) still marks *any* exception — a transient `httpx.RequestError`, a GitHub 503, or a permanent 4xx — as `status='failed'` immediately, same as before this PR. The `retry_count` column added here is only ever consulted by the crash-reclaim path, never by a job that ran and hit a transient error during execution. That's fine as an explicit scope decision, but since #22 is a single open issue covering all three, should this PR's merge close #22, or should retry/backoff become its own follow-up issue so it doesn't quietly fall off the radar? I'd lean toward the latter — can you open that as a separate issue and reference it here?

nazarli-shabnam (commented):
@Tarekchehahde 
Once CI is green, branch is up to date and that's settled, happy to take another look.

PR #123, opened 2026-07-11
fix(ui): guard token auto-resolve against stale cross-org responses (#97)

### LINKED ISSUE #97: fix(ui): owner-scoped token auto-resolve has no request-order guard, can bind wrong org's token
## Problem

Both `apps/ui/app/repos/[repo]/cache/page.tsx:33-40` and `apps/ui/app/security/page.tsx:91-96` run `resolveMutation.mutate(owner)` in a `useEffect` keyed on `owner`, and `onSuccess` unconditionally calls `setToken(data.token)` with no check that the response still corresponds to the current `owner`. `useMutation` does not cancel in-flight calls when `mutate` is invoked again.

## Failure scenario

A user types an owner name character by character (or switches quickly between two orgs) on the Security or Cache page; each keystroke past a valid length fires a new `tokens.resolve` request. If the response for an earlier/stale owner value arrives after the response for the current owner (plausible under normal network jitter), `setToken` is overwritten with a token belonging to the wrong organization. The user then unknowingly runs a scan or cache action with a different org's GitHub token silently populated in the field.

## Suggested fix

Guard `onSuccess` by comparing the resolved org against the current `owner` state (or the mutation variable) before calling `setToken`, or debounce/cancel via an AbortController per owner change.

## Severity

Medium — a security-relevant token mix-up between orgs, though it requires timing/race conditions to trigger.

Tarekchehahde:
## Summary
Fixes #97.

Token auto-resolve on security and cache pages ignores responses that no longer match the current owner.

## Test plan
- [x] `vitest run tests/lib/token-resolve.test.ts`
- [x] `cd apps/ui && bun run typecheck`

### DIFF
--- a/apps/ui/app/repos/[repo]/cache/page.tsx
+++ b/apps/ui/app/repos/[repo]/cache/page.tsx
@@ -14,6 +14,7 @@ import { Skeleton } from "@/components/ui/skeleton"
 import { AlertTriangle, Eye, KeyRound, Loader2, Trash2 } from "lucide-react"
 import { api } from "@/lib/api/client"
 import { parseOwnerRepo } from "@/lib/repo-segment"
+import { shouldApplyResolvedToken } from "@/lib/token-resolve"
 import { BarGroupChart } from "@/components/charts/bar-group-chart"
 import { CHART_COLORS } from "@/lib/charts/theme"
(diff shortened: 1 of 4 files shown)


nazarli-shabnam (dismissed):

Tarekchehahde:

nazarli-shabnam (approved):

PR #125, opened 2026-07-11
fix(checks): tolerate null security_and_analysis without crashing (#74)

### LINKED ISSUE #74: fix(checks): security_and_analysis: null crashes the entire check run
## Problem

`SecretScanningEnabled.run()` (`packages/checks/src/checks/github_checks.py:111-112`):

\`\`\`python
sec = repo.get("security_and_analysis", {})
if sec.get("secret_scanning", {}).get("status") == "enabled":
\`\`\`

The `{}` default only applies when the key is *missing* from the dict, not when GitHub returns it as JSON `null`. GitHub returns `"security_and_analysis": null` for repos where the calling token/user lacks admin access to view the security config (common with a limited-scope PAT). In that case `sec` is `None`, and `sec.get(...)` raises `AttributeError`.

## Failure scenario

An org admin runs analytics with a PAT that has repo access but not admin rights on every repo in the org. The first repo with `security_and_analysis: null` throws an unhandled `AttributeError` inside `run_all_checks`, which is not caught anywhere in `packages/checks/src/checks/runner.py` — the whole analytics request for the org fails with a 500 instead of returning results for the repos/checks that did succeed.

## Suggested fix

Use `repo.get("security_and_analysis") or {}` instead of the dict-default form, so an explicit `null` is also normalized to `{}`.

## Severity

High — a single repo's data shape can 500 the entire org's analytics.

Tarekchehahde:
## Summary
Fixes #74.

Treat explicit `security_and_analysis: null` as empty instead of calling `.get` on `None`.

## Test plan
- [x] `pytest packages/checks/tests/test_github_checks.py::test_secret_scanning_null_security_and_analysis_does_not_crash -v`

### DIFF
--- a/packages/checks/src/checks/github_checks.py
+++ b/packages/checks/src/checks/github_checks.py
@@ -113,7 +113,7 @@ class SecretScanningEnabled(Check):
         enabled = 0
         total = len(repos)
         for repo in repos:
-            sec = repo.get("security_and_analysis", {})
+            sec = repo.get("security_and_analysis") or {}
             if sec.get("secret_scanning", {}).get("status") == "enabled":
                 enabled += 1
         compliant = total > 0 and enabled == total
--- a/packages/checks/tests/test_github_checks.py
+++ b/packages/checks/tests/test_github_checks.py
@@ -2,7 +2,7 @@
 
 import pytest
 
-from checks.github_checks import OrgMFARequired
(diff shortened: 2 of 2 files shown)


Tarekchehahde:

nazarli-shabnam (approved):

PR #127, opened 2026-07-11
fix(checks): treat branch-protection 403/429 as unknown (#82)

### LINKED ISSUE #82: fix(checks): rate-limited/403 responses are reported as 'not protected'
## Problem

`BranchProtectionEnabled.run()` (`packages/checks/src/checks/github_checks.py:81-86`) catches `httpx.HTTPStatusError` per-repo and treats any such error as "not protected," with no distinction between a genuine 404 (branch protection not configured — a legitimate fail) and a 403 (rate-limited or the token lacks sufficient scope to see protection settings — not actually a compliance signal).

## Failure scenario

During a large org scan, GitHub secondary-rate-limits the token partway through. Repos hit after that point return 403 instead of the real protection status, and are recorded as "not protected" — producing a false security finding that makes the org look worse than it is, purely due to API throttling rather than actual configuration.

## Suggested fix

Distinguish status codes in the except handler: treat 404 as "not protected" (a real fail), but treat 403/429 as "unknown/unable to determine" and exclude those repos from the compliance denominator (or surface them separately) rather than counting them as failures.

## Severity

Medium — produces misleading security findings under normal, transient API conditions.

Tarekchehahde:
## Summary
Fixes #82.

403/429 responses when reading branch protection no longer score as unprotected.

## Test plan
- [x] `pytest packages/checks/tests/test_github_checks.py::test_branch_protection_rate_limit_counts_as_unknown -v`

### DIFF
--- a/packages/checks/src/checks/github_checks.py
+++ b/packages/checks/src/checks/github_checks.py
@@ -36,6 +36,15 @@ def _get_all_pages(base_url: str, path: str, token: str) -> list:
     return results
 
 
+def _branch_protection_status(exc: httpx.HTTPStatusError) -> str:
+    code = exc.response.status_code
+    if code == 404:
+        return "unprotected"
+    if code in (403, 429):
+        return "unknown"
+    return "unprotected"
+
+
 class OrgMFARequired(Check):
     metadata = CheckMetadata(
         check_id="organization_members_mfa_required",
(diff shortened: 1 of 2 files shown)


Tarekchehahde:

nazarli-shabnam (dismissed):
well done with the PRs. Fix the conflicts in here and its ready

Tarekchehahde:

nazarli-shabnam (approved):

PR #128, opened 2026-07-11
fix(checks): catch transient HTTPError in branch protection check (#99)

### LINKED ISSUE #99: fix(checks): BranchProtectionEnabled only catches HTTPStatusError, letting transient network errors abort the whole scan
## Problem

The per-repo branch-protection lookup in `packages/checks/src/checks/github_checks.py:81-86` wraps only `httpx.HTTPStatusError` in its `try/except`; any other `httpx` exception (`ConnectError`, `ReadTimeout`, `RemoteProtocolError`, etc.) propagates uncaught out of `run()`, and since `runner.run_all_checks` calls each check sequentially with no per-check isolation, one flaky network call on one repo kills the entire org's scan (including the already-computed MFA/secret-scanning results, which are discarded).

## Failure scenario

An org has 50 repos; on repo #30 the branch-protection request times out due to a transient network blip. `httpx.ReadTimeout` bubbles up out of `BranchProtectionEnabled.run`, out of `run_all_checks`, and the caller (API endpoint/worker) sees the whole scan fail with no partial results, even though 29 repos were already successfully checked and the other two checks would have succeeded.

## Suggested fix

Catch `httpx.HTTPError` (the common base class for both status and transport errors) around the per-repo branch call, and count unreachable branches the same way as inaccessible ones (or surface them as a distinct "errored" bucket).

## Severity

Medium — reliability issue that turns a single transient network error into a total scan failure.

Tarekchehahde:
## Summary
Fixes #99.

Count transient branch-protection network errors as `unknown` (same pattern as #82/#127) instead of reporting a false compliance failure.

Builds on #127 — rebased and aligned with the `unknown`/`error` status approach.

## Test plan
- [x] `pytest packages/checks/tests/test_github_checks.py::test_branch_protection_network_error_counts_as_unknown -v`

### DIFF
--- a/packages/checks/src/checks/github_checks.py
+++ b/packages/checks/src/checks/github_checks.py
@@ -100,6 +100,8 @@ class BranchProtectionEnabled(Check):
             except httpx.HTTPStatusError as exc:
                 if _branch_protection_status(exc) == "unknown":
                     unknown += 1
+            except httpx.HTTPError:
+                unknown += 1
         evaluable = checked - unknown
         if evaluable == 0:
             return {"status": "error", "value": {"checked": checked, "protected": protected, "unknown": unknown}}
--- a/packages/checks/src/checks/github_checks.py
+++ b/packages/checks/src/checks/github_checks.py
@@ -36,6 +36,15 @@ def _get_all_pages(base_url: str, path: str, token: str) -> list:
     return results
 
 
+def _branch_protection_status(exc: httpx.HTTPStatusError) -> str:
+    code = exc.response.status_code
+    if code == 404:
+        return "unprotected"
+    if code in (403, 429):
+        return "unknown"
+    return "unprotected"
+
+
 class OrgMFARequired(Check):
     metadata = CheckMetadata(
         check_id="organization_members_mfa_required",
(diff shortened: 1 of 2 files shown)


nazarli-shabnam (changes requested):
Hey @Tarekchehahde — this doesn't crash anything, so feel free to push back, but I want to flag the approach before it merges.

Right now a transient network error (timeout, connection drop) gets swallowed the same way a real "branch unprotected" result would — `checked` goes up, `protected` doesn't, so the check reports `fail` for a repo we actually just failed to check, not one that's actually unprotected.

#127 (same file) handles basically the same ambiguity by introducing an `unknown`/`error` status instead of guessing. Would be good to reuse that pattern here rather than silently mapping network errors to what looks like a compliance failure.

nazarli-shabnam:
Also please solve the conflicts in your prs and tag me again to review. Thanks beforehand.

Tarekchehahde:

nazarli-shabnam (dismissed):
Conflicts.

Tarekchehahde:

nazarli-shabnam (approved):

PR #134, opened 2026-07-12
fix(ui): track concurrent revoke rows with a Set (#129)

### LINKED ISSUE #129: fix(ui): rapid multi-row revoke clicks can re-enable a still-in-flight row
## Problem

PR #118 (closing #88) fixed the shared-mutation issue where every row's Revoke button disabled together, by tracking a single `revokingId` state (`apps/ui/app/settings/org/[login]/members/page.tsx`), set in `onMutate` and cleared in `onSettled`. Because `revokingId` only ever holds one id at a time, it can't represent more than one in-flight revoke.

## Failure scenario

An admin clicks Revoke on invitation A. Before A's request settles, they click Revoke on invitation B. `onMutate` overwrites `revokingId` from A's id to B's id, which re-enables row A's button even though A's DELETE request is still in flight. If the admin clicks A again in that window, a second concurrent revoke request fires for the same invitation id.

## Suggested fix

Track pending state as a `Set<number>` (or equivalent) of in-flight ids instead of a single id, adding on `onMutate` and removing on `onSettled` for that specific id, so multiple concurrent revokes are each tracked independently.

## Severity

Low — UX/double-submit edge case only; the revoke endpoint is presumably idempotent so a duplicate request likely no-ops, but the button state briefly lies about a row's in-flight status.

Tarekchehahde:
## Summary
- Replace single `revokingId` state with a `Set<number>` so each invitation row tracks its own in-flight revoke independently
- Extract small `revoke-pending` helpers and add a unit test for the concurrent-revoke scenario described in the issue

Fixes #129

## Test plan
- [x] `bun run test tests/lib/revoke-pending.test.ts` (pass)
- [x] `bun run typecheck` (pass)
- [ ] Manual: org Members page — revoke two pending invites quickly; both buttons stay disabled until each request settles

### DIFF
--- a/apps/ui/app/settings/org/[login]/members/page.tsx
+++ b/apps/ui/app/settings/org/[login]/members/page.tsx
@@ -8,6 +8,7 @@ import { Input } from "@/components/ui/input"
 import { Button } from "@/components/ui/button"
 import { Loader2, Mail, X } from "lucide-react"
 import { api } from "@/lib/api/client"
+import { addRevokingId, isRevoking, removeRevokingId } from "@/lib/revoke-pending"
 import type { InvitationOut } from "@/lib/api/types"
 
 export default function OrgMembersPage() {
@@ -22,7 +23,7 @@ export default function OrgMembersPage() {
     queryFn: () => api.invitations.list(orgLogin),
   })
 
-  const [revokingId, setRevokingId] = useState<number | null>(null)
+  const [revokingIds, setRevokingIds] = useState<Set<number>>(() => new Set())
 
   const invite = useMutation({
     mutationFn: () => api.invitations.create(orgLogin, email.trim()),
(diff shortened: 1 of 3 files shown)


nazarli-shabnam:
@Tarekchehahde Hi again! Mind adding your name mark to commits?

Tarekchehahde:
@nazarli-shabnam Rebuilt branch on latest `main` (was stale). Added Signed-off-by: Tarek Chehahde on commit. Ready for re-review.

nazarli-shabnam (approved):
Verified against #129: revoke state is now a `Set<number>` added in `onMutate` and removed in `onSettled` for that specific invitation id, so concurrent revokes on different rows no longer clobber each other's disabled state. Clean helper module with unit tests covering the independent-tracking case, CI green. Thanks!

nazarli-shabnam (approved):

PR #148, opened 2026-07-12
fix(api): verify installation_id against GitHub on sync (#141)

### LINKED ISSUE #141: fix(api): installation-sync endpoints accept client-supplied installation_id with no server-side verification
## Problem

`POST /orgs/{org_login}/installations/sync` and `POST /me/installations/sync` (`apps/api/src/routers/installations.py`) accept a client-supplied `installation_id`/`account_login`/`auth_mode` in the request body and persist them directly via `installation_repo.create`/`upsert` — there is no call to GitHub's Installations API (`GET /orgs/{org}/installation` or `GET /users/{user}/installation`) to confirm the `installation_id` actually exists, belongs to that account, or that the caller has any real relationship to it.

`installation_repo.create` also sets `token_ref = f"tok_{account_login}"` — a symbolic placeholder string, not a real credential reference (per the existing comment on `GitHubInstallation.token_ref` in `core/db.py`).

## Failure scenario

Today this is low-risk in isolation, because nothing downstream reads `token_ref` to mint or use a real token (see the related issue: `get_installation_token()` is never called). But the moment installation-token auth gets wired into an actual feature, this becomes a real integrity gap: any authenticated user/org-admin can register an arbitrary or fabricated `installation_id` for their org/account with zero server-side proof it's real, and the system would treat it as legitimate.

## Suggested fix

Before (or as part of) wiring installation-token auth into real features, add server-side verification in the sync endpoints: call GitHub's installation-lookup API with the App's own credentials and confirm the `installation_id` genuinely belongs to the target org/account before persisting it.

## Severity

Medium — no active exploit today (nothing consumes the unverified data yet), but a real trust boundary is missing, and it's easy to forget to add this check later once installation-token auth actually ships.

Tarekchehahde:
## Summary
Fixes #141.

Verify client-supplied installation_id against GitHub Installations API before persisting sync rows.

## Test plan
- [x] pytest apps/api/tests/test_installation_verify.py (3 passed)

### DIFF
--- a/apps/api/src/routers/installations.py
+++ b/apps/api/src/routers/installations.py
@@ -24,6 +24,7 @@ from src.schemas.installation import (
     SyncInstallationsInput,
     SyncInstallationsResponse,
 )
+from src.services.installation_verify import verify_sync_payload
 
 router = APIRouter()
 
@@ -43,6 +44,7 @@ def sync_org_installation(
     db: Session = Depends(get_db),
 ):
     assert_owner_matches_org(payload.account_login, ctx)
+    verify_sync_payload(payload)
     row = installation_repo.create(
         db,
         account_login=payload.account_login,
@@ -84,6 +86,7 @@ def sync_personal_installation(
             status_code=status.HTTP_403_FORBIDDEN,
             detail="account_login must match your own GitHub account",
         )
+    verify_sync_payload(payload)
     row = installation_repo.create(
         db,
         account_login=payload.account_login,
--- a/apps/api/src/services/installation_verify.py
+++ b/apps/api/src/services/installation_verify.py
@@ -1,0 +1,56 @@
+"""Verify GitHub App installation metadata against client sync payloads."""
+
+from __future__ import annotations
+
+import httpx
+from fastapi import HTTPException, status
+
+from src.core.config import settings
+from src.schemas.installation import SyncInstallationsInput
+from src.services import github_app
+
+
+def fetch_installation(installation_id: int) -> dict:
+    """Load installation metadata from GitHub using an App JWT."""
+    app_jwt = github_app.generate_app_jwt()
+    url = f"{settings.github_api_base}/app/installations/{installation_id}"
+    headers = {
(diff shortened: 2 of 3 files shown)


nazarli-shabnam (changes requested):
The verification logic itself (calling GitHub's `/app/installations/{id}` with the App JWT and checking `account.login`/`account.type` against the payload) is the right approach and matches what #141 asked for, and the new unit tests for `installation_verify.py` look solid.

But this breaks CI — `Python Tests + Compile` is failing:

```
FAILED apps/api/tests/test_installations.py::test_sync_org_installation_admin_ok - assert 503 == 200
FAILED apps/api/tests/test_installations.py::test_sync_personal_installation - assert 503 == 200
FAILED apps/api/tests/test_installations.py::test_sync_org_installation_upserts_existing_row - AssertionError: assert 503 == 200
3 failed, 168 passed
```

Since `auth_mode` defaults to `"app"` (`SyncInstallationsInput.auth_mode: str = "app"`) and CI doesn't have `GITHUB_APP_PRIVATE_KEY`/`GITHUB_APP_ID` configured, `verify_sync_payload` now hits `github_app.GitHubAppNotConfigured` on every existing sync call that doesn't explicitly mock it, turning into a 503. The pre-existing `test_installations.py` suite wasn't updated to account for the new verification step, so these three previously-green tests now fail for real.

That also points at a production-facing gap, not just a test one: per the comment in `config.py`, the GitHub App is "optional until registered" — any deployment that hasn't registered a GitHub App yet (but is still using `auth_mode: "app"` as the schema default) will now get a hard `503` on every installation sync, where before this it worked. Worth deciding explicitly whether that's the intended behavior (installation sync requires the App to be configured now, full stop) or whether unconfigured-App deployments need a distinct code path — either way the existing tests need to reflect the decision.

Can you update `test_installations.py` to mock `verify_sync_payload`/`fetch_installation` where appropriate (or adjust the auth_mode used in those fixtures) so CI is green, and confirm the unconfigured-App behavior is intentional?

PR #149, opened 2026-07-12
fix(ui): redirect Collaborators to working members page (#138)

### LINKED ISSUE #138: feat(ui): Collaborators page is a dead stub while a working member-management page exists, unlinked
## Problem

`apps/ui/app/collaborators/page.tsx` renders:

```tsx
<EmptyState
  icon={Users}
  title="Collaborators coming soon"
  description="View your organization roster, manage pending invitations, and fix member permissions from a single dashboard."
/>
```

But `/settings/org/[login]/members` already implements exactly this — real invite creation, listing, and revocation (`api.invitations.create/list/revoke`), fully tested. The sidebar's "Collaborators" nav item (`apps/ui/components/app-sidebar.tsx`) routes to the dead stub, not the working page — and there's no link from the stub to the real one either.

## User impact

A user clicks the most obviously-named nav item for member management, hits a "coming soon" dead end, and never discovers the real feature exists at all (it's only reachable via Settings → an org's "Manage members" link).

## Suggested fix

Either:
- (a) Make `/collaborators` redirect to (or embed) the real `/settings/org/[login]/members` page for the user's default/selected org, or
- (b) Remove the `/collaborators` nav item entirely and surface "Manage members" more prominently from the main nav instead of burying it in Settings.

Either way, users should not see a dead stub for a feature that's already built and working elsewhere.

## Severity

Medium — discoverability/UX gap for an already-working feature, not a missing feature.

Tarekchehahde:
## Summary
Fixes #138.

Redirect org admins from /collaborators to /settings/org/{login}/members; others get a Settings link.

## Test plan
- [ ] Manual: visit /collaborators as org admin

### DIFF
--- a/apps/ui/app/collaborators/page.tsx
+++ b/apps/ui/app/collaborators/page.tsx
@@ -1,23 +1,46 @@
-export const metadata = { title: "Collaborators · clevis" }
+"use client"
 
+import { useEffect } from "react"
+import { useRouter } from "next/navigation"
+import { useQuery } from "@tanstack/react-query"
+import { Loader2 } from "lucide-react"
 import { PageHeader } from "@/components/page-header"
-import { EmptyState } from "@/components/empty-state"
-import { Users } from "lucide-react"
+import { EmptyStatePage } from "@/components/empty-state"
+import { api } from "@/lib/api/client"
 
 export default function CollaboratorsPage() {
+  const router = useRouter()
+  const { data: orgs, isLoading } = useQuery({
+    queryKey: ["orgs", "mine"],
+    queryFn: () => api.orgs.mine(),
+  })
+
+  const adminOrg = orgs?.find((o) => o.role === "admin")
+
+  useEffect(() => {
+    if (adminOrg) {
+      router.replace(`/settings/org/${adminOrg.github_login}/members`)
+    }
+  }, [adminOrg, router])
+
+  if (isLoading || adminOrg) {
+    return (
+      <>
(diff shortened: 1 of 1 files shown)


nazarli-shabnam:
@Tarekchehahde Hey. Fix ci job first. and i'd suggest to link your github account to your code editor.

nazarli-shabnam (changes requested):
@Tarekchehahde Same issue as #151 (and it's the identical bug, so flagging it the same way): `adminOrg.github_login` is used to build the redirect target, but `api.orgs.mine()` returns `MyOrgMembership[]` whose field is `org_login`, not `github_login` — there's no `github_login` field on that type. So the redirect is to `/settings/org/undefined/members`, which doesn't actually solve #138 (users still hit a dead end, just via a broken redirect instead of a static stub). CI confirms this — Typecheck/Build, E2E, and Docker Build Verification are all failing on this PR.

Also: this hardcodes redirecting to the first *admin* org (`orgs?.find(o => o.role === "admin")`), with no fallback for a non-admin member who wants to just view the roster — they'll sit on the loading/empty state forever since `adminOrg` will never resolve for them. The issue's suggested fix (a) says redirect "for the user's default/selected org" — worth deciding whether non-admin users should still land somewhere useful (e.g., the members page in read-only mode, if that page supports it) rather than a permanent "open Settings" empty state.

Can you fix the field name and repush so CI passes, and let me know the intended behavior for non-admin org members?

nazarli-shabnam (commented):
Also update the branch please

PR #151, opened 2026-07-12
fix(ui): wire sidebar Invite members to org members page (#139)

### LINKED ISSUE #139: feat(ui): sidebar 'Invite members' button is a disabled dead placeholder duplicating a working feature
## Problem

`apps/ui/components/app-sidebar.tsx`'s profile dropdown (`ProfileDropdown` component) has an "Invite members" button:

```tsx
<button
  disabled
  className="... disabled:opacity-50 disabled:cursor-not-allowed"
  title="Coming soon"
>
  <UserPlus className="size-3" />
  Invite members
</button>
```

It's permanently disabled with a "Coming soon" tooltip, styled identically to the working "Settings" button right next to it. The real invite flow already exists and works at `/settings/org/[login]/members`.

## User impact

A user opens the profile dropdown, sees an "Invite members" action that looks real (same styling as "Settings"), tries to click it, and nothing happens except a tooltip they may not even notice hovering over a disabled button. They have no way to discover the actual invite flow lives elsewhere.

## Suggested fix

Wire this button to navigate to `/settings/org/[login]/members` (using the user's current/default org context, same as the working `default_org` pattern used on the Security page), or remove it from the dropdown entirely if a duplicate entry point isn't wanted.

## Severity

Low-Medium — same underlying issue as the Collaborators stub (#137 or nearby) — a real feature exists, but the most obvious entry point to it is dead.

Tarekchehahde:
## Summary
Fixes #139.

Link org admins to the working invite flow instead of a disabled placeholder.

## Test plan
- [ ] Manual: profile dropdown Invite members

### DIFF
--- a/apps/ui/components/app-sidebar.tsx
+++ b/apps/ui/components/app-sidebar.tsx
@@ -2,6 +2,7 @@
 
 import { usePathname, useRouter } from "next/navigation"
 import { useEffect, useRef, useState } from "react"
+import { useQuery } from "@tanstack/react-query"
 import Link from "next/link"
 import { Settings, Check, LogOut, UserPlus } from "lucide-react"
 import {
@@ -16,6 +17,7 @@ import {
   SidebarSeparator,
 } from "@/components/ui/sidebar"
 import { useAuth } from "@/lib/auth-context"
+import { api } from "@/lib/api/client"
 
 // Settings is no longer in the sidebar nav — it lives inside the profile dropdown.
 const groups = [
@@ -50,10 +52,12 @@ interface Profile {
 
 function ProfileDropdown({
   profile,
+  inviteHref,
   onClose,
   onSignOut,
 }: {
   profile: Profile
+  inviteHref?: string
   onClose: () => void
   onSignOut: () => void
 }) {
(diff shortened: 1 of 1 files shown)


nazarli-shabnam (changes requested):
CI is red on this one (UI Typecheck + Test + Build, E2E Tests, Docker Build Verification all failing), and it's a real bug, not flakiness:

`app-sidebar.tsx` builds the invite link from `adminOrg.github_login`, but `api.orgs.mine()` returns `MyOrgMembership[]` (`apps/ui/lib/api/types.ts`), whose field is `org_login`, not `github_login`. There's no `github_login` field on that type anywhere. So `adminOrg.github_login` is always `undefined`, and the rendered link is literally `/settings/org/undefined/members` — this doesn't fix #139, it replaces one dead button with a differently-broken one. That's exactly what's failing typecheck.

Separate from the typo: `app-sidebar.tsx` already has a `defaultOrg` value read from `localStorage.getItem("default_org")` a few lines below (used for the `profile.org` display right in this same dropdown). This PR ignores that and instead does `orgs?.find(o => o.role === "admin")`, which — once the field name is fixed — will pick an arbitrary admin org, not necessarily the one shown as the user's current org in the dropdown header. For a user who's admin of multiple orgs, "Invite members" could silently navigate to a different org than the one they think they're looking at. The issue itself points at "using the user's current/default org context, same as the working `default_org` pattern used on the Security page" — worth using that instead of (or as a fallback/preference over) the first admin match.

nazarli-shabnam (commented):
Can you fix the field name (`org_login`) and align the org selection with the existing `default_org` pattern, then repush so CI is green?
@Tarekchehahde

PR #152, opened 2026-07-12
fix(api): add GitHub App webhook endpoint (#142)

### LINKED ISSUE #142: fix(api): no webhook-receiving endpoint exists despite GITHUB_APP_WEBHOOK_SECRET being configured
## Problem

`apps/api/src/core/config.py` defines `github_app_webhook_secret: SecretStr | None` as a setting, implying GitHub App webhook events (installation created/deleted, permissions changed, etc.) are meant to be received and verified. But `apps/api/src/main.py`'s full router list (`health, auth, github_auth, installations, orgs, invitations, analytics, actions_cache, jobs, audit, tokens, config`) has no webhook-receiving router, and a repo-wide search for "webhook" only turns up this one unused config field.

## User impact

Without webhook handling, the `github_installations` table (see related issues about unverified `installation_id` and unused installation tokens) can never be kept in sync with reality — if a user uninstalls the GitHub App, revokes access, or an installation's permissions change, clevis has no way to find out and will keep treating stale/invalid installation rows as valid indefinitely.

## Suggested fix

Add a `POST /webhooks/github` endpoint that:
1. Verifies the `X-Hub-Signature-256` header against `GITHUB_APP_WEBHOOK_SECRET` (reject if missing/invalid).
2. Handles at minimum `installation` (created/deleted) and `installation_repositories` events to keep `github_installations` rows in sync.

This is a prerequisite for the installation-token auth path (separate issue) to be trustworthy long-term, not just at sync time.

## Severity

Low-Medium — no active bug, but blocks the installation-based auth story from ever being safe to fully rely on.

Tarekchehahde:
## Summary
Fixes #142.

POST /webhooks/github verifies X-Hub-Signature-256 and removes installations on deleted events.

## Test plan
- [x] pytest apps/api/tests/test_webhooks.py (2 passed)

### DIFF
--- a/apps/api/src/main.py
+++ b/apps/api/src/main.py
@@ -19,6 +19,7 @@ from src.routers import (
     jobs,
     orgs,
     tokens,
+    webhooks,
 )
 
 # CORS allowed origins are a deploy-time security boundary, set via the CORS_ORIGINS env var.
@@ -65,3 +66,4 @@ app.include_router(jobs.router, prefix="/jobs", tags=["jobs"])
 app.include_router(audit.router, prefix="/audit", tags=["audit"])
 app.include_router(tokens.router, prefix="/tokens", tags=["tokens"])
 app.include_router(config.router, prefix="/config", tags=["config"])
+app.include_router(webhooks.router, tags=["webhooks"])
--- a/apps/api/src/repositories/installation_repo.py
+++ b/apps/api/src/repositories/installation_repo.py
@@ -83,3 +83,8 @@ def list_for_user(db: Session, owner_user_id: int) -> list[GitHubInstallation]:
         .order_by(GitHubInstallation.created_at.desc())
         .all()
     )
+
+
+def delete_by_installation_id(db: Session, installation_id: int) -> None:
+    db.query(GitHubInstallation).filter(GitHubInstallation.installation_id == installation_id).delete()
+    db.commit()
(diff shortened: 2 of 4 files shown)


nazarli-shabnam (changes requested):
The signature verification itself looks correct (constant-time comparison, rejects missing/invalid `X-Hub-Signature-256`, 503s cleanly if the secret isn't configured), and the `installation` → `deleted` → row-removal path is tested and works. But this only covers 1 of the 3 events #142 explicitly asks for "at minimum":

- `installation` **created** — not handled at all (only `deleted` is).
- `installation_repositories` — not handled at all; any event other than `installation`/`deleted` just falls through to a 204 no-op.

I can see an argument that `created` is redundant with the existing interactive install-callback flow (`installation_repo.upsert`), and that `installation_repositories` may not matter given `github_installations` doesn't track repo-level scope today — but the PR doesn't make that case. Compare with #159, which took a narrower path than its issue suggested but explicitly justified the trade-off in the description. This PR's description ("removes installations on deleted events") doesn't explain why the other two events from the issue were dropped, so I can't tell if it's a considered scope decision or a gap.

Also worth calling out since this is new unauthenticated attack surface (verified only by HMAC, no RBAC) touching installation/auth data — per AGENTS.md, PRs touching auth-adjacent code should flag that explicitly in the description; this one doesn't.

nazarli-shabnam (commented):
Could you either add installation(created)/installation_repositories handling, or update the description to explicitly justify why deleted alone satisfies https://github.com/nazarli-shabnam/clevis/issues/142 and flag this as new auth-adjacent surface? Happy to re-review quickly once that's in. @Tarekchehahde

PR #153, opened 2026-07-12
fix(worker): dispatch queued jobs by job_type (#143)

### LINKED ISSUE #143: fix(worker): worker only dispatches one hardcoded job_type; any other job type queues forever
## Problem

`apps/worker/src/worker.py`'s poll query hardcodes the job type it looks for:

```sql
WHERE status = 'queued' AND job_type = 'github.clear_actions_cache'
```

But `job_repo.enqueue(db, job_type, payload)` (`apps/api/src/repositories/job_repo.py`) takes an arbitrary `job_type` string, and the `jobs` table schema itself is fully generic (composite index on `(status, job_type)`, per the architecture doc). Only `github.clear_actions_cache` is ever actually enqueued today (from `cache_service.clear`), and only that literal string is ever picked up.

## Failure scenario

Any future feature that calls `job_repo.enqueue()` with a different `job_type` (e.g. a bulk security scan, a scheduled report) would insert a row with `status='queued'` that the worker's poll query will never match. The job would sit in the queue forever with no error, no timeout, no visibility that anything is wrong — it just silently never runs.

## Suggested fix

Replace the single hardcoded `job_type` filter with a dispatch table/mapping (`job_type -> handler function`) so the worker's poll query matches any known job type and routes to the appropriate handler, falling back to marking genuinely unknown job types as `failed` with a clear reason (rather than looping forever or crashing).

## Severity

Medium — not an active bug (nothing enqueues other job types today), but a structural trap for the next feature that uses the job queue as designed.

Tarekchehahde:
## Summary
Fixes #143.

Poll any queued job and route via handler map; unknown types fail clearly.

## Test plan
- [x] pytest apps/worker/tests/test_process_job.py (6 passed)

### DIFF
--- a/apps/worker/src/worker.py
+++ b/apps/worker/src/worker.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import time
+from collections.abc import Callable
 
 import httpx
 import psycopg
@@ -17,6 +18,8 @@ log = logging.getLogger(__name__)
 
 # psycopg.connect() expects plain postgresql://, not the SQLAlchemy +psycopg dialect prefix
 _DB_URL = settings.database_url.get_secret_value().replace("postgresql+psycopg://", "postgresql://")
+
+JOB_TYPE_CLEAR_ACTIONS_CACHE = "github.clear_actions_cache"
 
 
 def _read_app_config(key: str, default: str) -> str:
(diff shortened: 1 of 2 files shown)


nazarli-shabnam (approved):
Verified against #143: the poll query no longer hardcodes `job_type = 'github.clear_actions_cache'`, dispatch now goes through a `job_type -> handler` map, and unknown job types are marked `failed` with a clear reason instead of looping forever — exactly what the issue asked for. Existing tests updated for the new `process_job` signature and a new test covers the unknown-job_type path. No migration/RBAC/crypto touched. Thanks!

PR #154, opened 2026-07-12
fix(checks): support personal-account security scans (#144)

### LINKED ISSUE #144: fix(api): personal-account security scan likely errors for real personal (non-org) GitHub accounts
## Problem

`POST /me/analytics/overview` (`apps/api/src/routers/analytics.py`) lets any authenticated user run a security scan against `owner` = their own personal GitHub login, with no org membership required. But the checks it runs underneath (`checks.runner.run_all_checks` → `OrgMFARequired.run()` calling `GET /orgs/{owner}`, and repo listing via `GET /orgs/{owner}/repos` in `packages/checks/src/checks/runner.py`) are **org-only** GitHub REST endpoints — they 404 for a personal (User-type) account.

## Failure scenario

A solo user with no GitHub org signs up, connects their personal account, and clicks "Run scan" on the Security page (`/security`, which calls `api.analytics.overview(owner, token)` with the user's own personal login as `owner`). The underlying GitHub API calls 404 instead of returning real data — the user sees an error (or a check reporting `"error"` status, per `OrgMFARequired`'s own error-handling) instead of a working personal security scan. This looks like a route that was added for personal accounts without adapting the check logic, which was built org-first.

## Suggested fix

Either:
- (a) Adapt the checks (or add personal-account variants) to use the correct user-scoped GitHub endpoints (`GET /user`, `GET /users/{username}/repos`, etc.) when running against a personal account rather than an org, or
- (b) If personal-account scanning genuinely isn't supported yet, have `/me/analytics/overview` return a clear "personal accounts aren't supported yet, connect an organization" response instead of letting it hit org-only endpoints and fail unclearly.

## Severity

Medium-High — a core feature (security scanning) appears broken for an entire class of real users (individuals without an org), with no clear indication of why.

Tarekchehahde:
## Summary
Fixes #144.

Use /users/{owner}/repos for personal accounts; org MFA check is not_applicable for users.

## Test plan
- [x] pytest packages/checks/tests/test_runner_account_type.py (1 passed)

### DIFF
--- a/packages/checks/src/checks/runner.py
+++ b/packages/checks/src/checks/runner.py
@@ -1,24 +1,52 @@
 import logging
 
+import httpx
+
 from checks.github_checks import BranchProtectionEnabled, OrgMFARequired, SecretScanningEnabled, _get_all_pages
 
 logger = logging.getLogger(__name__)
 
 
+def _resolve_account(base_url: str, owner: str, token: str) -> tuple[str, str]:
+    """Return (repos_path, account_type) for owner login."""
+    headers = {
+        "Authorization": f"Bearer {token}",
+        "Accept": "application/vnd.github+json",
+        "X-GitHub-Api-Version": "2022-11-28",
+    }
+    with httpx.Client(timeout=20) as client:
+        user_resp = client.get(f"{base_url}/users/{owner}", headers=headers)
+        if user_resp.status_code == 200:
+            data = user_resp.json()
+            if data.get("type") == "User":
+                return f"/users/{owner}/repos", "User"
+        org_resp = client.get(f"{base_url}/orgs/{owner}", headers=headers)
+        if org_resp.status_code == 200:
+            return f"/orgs/{owner}/repos", "Organization"
+    raise RuntimeError(f"Could not resolve GitHub account for owner {owner!r}")
+
+
 def run_all_checks(owner: str, token: str, base_url: str = "https://api.github.com") -> dict:
-    repos = _get_all_pages(base_url, f"/orgs/{owner}/repos", token)
+    repos_path, account_type = _resolve_account(base_url, owner, token)
+    repos = _get_all_pages(base_url, repos_path, token)
 
     checks = [OrgMFARequired(), BranchProtectionEnabled(), SecretScanningEnabled()]
     results = []
     for check in checks:
-        try:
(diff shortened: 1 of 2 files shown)


nazarli-shabnam (changes requested):
Thanks for taking this on, but I don't think this should merge as-is.

Issue #144 was already fixed and closed by #159 (merged 2026-07-12T19:41), which took the *other* option listed in the issue: `/me/analytics/overview` now resolves the account type in `analytics.py`/`analytics_service.get_account_type()` and returns a clear `422` before ever calling `get_overview()`/`run_all_checks()` when `owner` is a personal (`User`-type) account.

That means this PR's change to `packages/checks/src/checks/runner.py` (`_resolve_account`, the `not_applicable` branch for `organization_members_mfa_required`) is now **dead code** on the only caller that matters (`/me/analytics/overview`) — the router rejects personal accounts before `run_all_checks` is ever invoked with a User-type owner. It also duplicates the `GET /users/{owner}` lookup #159 already does, just one layer deeper, with no caller left that would exercise the `User` branch here.

Two separate concerns worth flagging on top of the redundancy:
- `_resolve_account` swallows a non-200/404 from GitHub (rate limit, 5xx) silently and falls through to the `orgs` call, then raises a generic `RuntimeError` if both fail — that's a behavior change for the org-scoped `/orgs/{org_login}/analytics/overview` path too, since `run_all_checks` is shared by both routers. Previously a transient GitHub error there would propagate normally; now it's masked behind a new exception type the callers don't handle.
- No test covers the `Organization` branch of `_resolve_account` (only `User` is tested, and only by mocking `_resolve_account` itself away — `test_runner_account_type.py` never actually exercises the two live `httpx` calls or their error paths).

nazarli-shabnam (commented):
@Tarekchehahde 
Given https://github.com/nazarli-shabnam/clevis/issues/144 is closed and https://github.com/nazarli-shabnam/clevis/pull/159 already ships the maintainer-chosen fix, could you either close this as superseded, or if you still want checks-level personal-account support as a genuine follow-up (separate from https://github.com/nazarli-shabnam/clevis/issues/144), open a fresh issue for it and rebase this against the current run_all_checks/_run_overview flow so it isn't unreachable code?

PR #159, opened 2026-07-12
fix(api): reject personal-account scans on /me/analytics/overview with a clear error

### LINKED ISSUE #144: fix(api): personal-account security scan likely errors for real personal (non-org) GitHub accounts

gitstatic-collab:
## What changed and why

`POST /me/analytics/overview` let any authenticated user run a security scan
against an arbitrary `owner` login, with no check that the login was actually
a GitHub *organization*. The checks underneath — `OrgMFARequired` (calls
`GET /orgs/{owner}`) and the repo-listing call shared by all three checks
(`GET /orgs/{owner}/repos`) — are org-only GitHub REST endpoints. A solo user
with no org, scanning their own personal login, got an opaque
`400 GitHub API error: 404` instead of any explanation of what went wrong.

This adds an account-type check (`GET /users/{owner}`) before running the
scan on the `/me/analytics/overview` path. If the login resolves to a
personal (`User`-type) account, the endpoint now returns a `422` with a clear
message ("Personal GitHub accounts aren't supported for security scanning
yet. Connect a GitHub organization instead.") instead of letting the request
fail unclearly against org-only endpoints.

I chose this minimal guard over adapting the checks themselves to support
personal accounts (issue #144's other suggested option) because that's a
materially larger change touching the shared `packages/checks` package
(`OrgMFARequired` in particular has no real per-user equivalent — GitHub
doesn't expose "does this personal account require 2FA" the way it does for
orgs) and felt out of scope for this fix. Full personal-account support can
be a follow-up.

The `/orgs/{org_login}/analytics/overview` path is
`assert_owner_matches_org` already ties `owner` to a real org there, so it
can't hit this failure mode.

No migration, no RBAC/auth/token-encryption change

Fixes #144

### DIFF
--- a/apps/api/src/routers/analytics.py
+++ b/apps/api/src/routers/analytics.py
@@ -9,7 +9,7 @@ from src.core.auth import UserOut, require_auth
 from src.core.db import get_db
 from src.core.rbac import OrgContext, assert_owner_matches_org, require_org_role
 from src.schemas.analytics import AnalyticsInput, AnalyticsResponse
-from src.services.analytics_service import get_overview
+from src.services.analytics_service import get_account_type, get_overview
 from src.services.token_resolution import NoGitHubTokenAvailable, resolve_org_token, resolve_personal_token
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


nazarli-shabnam (approved):
Thanks for the contribution!

PR #171, opened 2026-07-16
fix(api): verify installation_id against GitHub before persisting sync rows (#141)

### LINKED ISSUE #141: fix(api): installation-sync endpoints accept client-supplied installation_id with no server-side verification

nazarli-shabnam:
## ⚠️ Touches auth/trust-boundary code

This PR changes `apps/api/src/routers/installations.py` and `apps/api/src/services/github_app.py` — the installation-sync auth flow. Flagging per AGENTS.md guardrail #3/#2. No RBAC role requirements changed, no token-encryption logic touched, no migration — this only adds a verification call before data is trusted and persisted; it doesn't relax anything.

## Summary

`POST /orgs/{org_login}/installations/sync` and `POST /me/installations/sync` accepted a client-supplied `installation_id`/`account_login`/`auth_mode` and persisted them directly via `installation_repo.create`/`upsert` — there was no call to GitHub's Installations API to confirm the `installation_id` actually exists or belongs to the claimed account. Today this is low-risk in isolation (nothing downstream reads `token_ref` to mint a real token yet), but any authenticated user/org-admin could register an arbitrary or fabricated `installation_id` for their org/account with zero server-side proof it's real — a real trust-boundary gap that's easy to forget once installation-token auth actually ships and starts consuming this data.

Changes:
- `github_app.py`: new `get_installation(installation_id) -> dict`, calling `GET /app/installations/{id}` with the App's own JWT (not an installation token — we don't have one for an unverified ID, that's the whole point). Raises `httpx.HTTPStatusError` (404 if it doesn't exist) or `GitHubAppNotConfigured`.
- `installations.py`: new `_verify_installation(installation_id, account_login, account_type)`, called from both sync endpoints right after their existing ownership checks (`assert_owner_matches_org` / the `github_login` match) and before `installation_repo.create`. It fetches the installation and compares `account.login` (case-insensitive) and `account.type` against what the client claimed:
  - Mismatch or nonexistent `installation_id` → `422`, nothing persisted.
  - GitHub App not configured → `503` — silently accepting unverified data in this case would defeat the entire point of the fix, so verification is required, not best-effort.
  - Other GitHub API errors → `400`; network errors → `503` (matching the existing pattern in `analytics.py`).
  - `installation_id` omitted entirely → verification is skipped (nothing to verify), matching the schema's existing optional design.

Closes #141

## Checks

- [ ] `cd apps/ui && bun run typecheck && bun run build` (if UI changed) — not applicable, no UI changes
- [x] `python -m compileall apps/api/src apps/worker/src`
- [x] Relevant `pytest` tests pass
- [ ] Docker images still build, if `Dockerfile`/deps changed — not applicable, no deps changed
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)

Updated existing `test_installations.py` sync tests to mock the new verification call, and added tests for: mismatched account rejected (422, nothing persisted), nonexistent `installation_id` rejected (422), App-not-configured returns 503, and verification skipped when `installation_id` is omitted. Also added `test_get_installation_uses_app_jwt_not_installation_token` (decodes the auth header to confirm it's actually the App JWT, not an installation token) and a 404 test in `test_github_app.py`. Full `pytest -q` suite (202 tests) passes against a real Postgres instance.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y7RDvvVu5mqC4pybwNqq7A)_

### DIFF
--- a/apps/api/src/routers/installations.py
+++ b/apps/api/src/routers/installations.py
@@ -12,6 +12,7 @@
   POST /me/installations/sync                connect a personal (User-type) GitHub installation
 """
 
+import httpx
 from fastapi import APIRouter, Depends, HTTPException, status
 from sqlalchemy.orm import Session
 
@@ -24,8 +25,42 @@ from src.schemas.installation import (
     SyncInstallationsInput,
     SyncInstallationsResponse,
 )
+from src.services import github_app
 
 router = APIRouter()
+
+
+def _verify_installation(installation_id: int | None, account_login: str, account_type: str) -> None:
+    """Confirm a client-supplied installation_id genuinely belongs to the claimed
+    account before it's persisted as trusted data. No-op if installation_id wasn't
+    supplied — there's nothing to verify in that case."""
+    if installation_id is None:
+        return
+    try:
+        installation = github_app.get_installation(installation_id)
+    except github_app.GitHubAppNotConfigured:
+        raise HTTPException(
+            status_code=503,
+            detail="GitHub App is not configured; cannot verify installation_id",
+        )
+    except httpx.HTTPStatusError as exc:
+        if exc.response.status_code == 404:
+            raise HTTPException(status_code=422, detail=f"installation_id {installation_id} does not exist")
+        raise HTTPException(status_code=400, detail=f"GitHub API error: {exc.response.status_code}")
+    except httpx.RequestError:
+        raise HTTPException(status_code=503, detail="GitHub API unreachable")
+
+    account = installation.get("account") or {}
+    actual_login = account.get("login", "")
+    actual_type = account.get("type", "")
+    if actual_login.lower() != account_login.lower() or actual_type != account_type:
+        raise HTTPException(
+            status_code=422,
+            detail=(
+                f"installation_id {installation_id} belongs to {actual_type} "
+                f"'{actual_login}', not {account_type} '{account_login}'"
+            ),
+        )
 
 
 @router.get("/orgs/{org_login}/installations", response_model=list[InstallationOut])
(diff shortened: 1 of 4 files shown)


nazarli-shabnam (commented):
installation sync verify

Issue #186, opened 2026-07-17
fix(ui): setupRequired() has no .catch(), unreachable API silently swallowed


nazarli-shabnam:
## Problem

`api.auth.setupRequired()` is called with no `.catch()` in both `apps/ui/app/login/page.tsx` and `apps/ui/app/setup/page.tsx`. This is called out as known debt in `docs/plan.md:241`: "if the API is unreachable on page load the redirect is silently swallowed."

## Failure scenario

If the API is temporarily unreachable (deploy in progress, network blip) when a user lands on `/login` or `/setup`, the `setupRequired()` call rejects, the `.then()`-based redirect logic never runs, and the page silently shows nothing useful — no error message, no retry, no indication of what went wrong.

## Suggested fix

Add a `.catch()` (or wrap in try/catch if using async/await) on both call sites that surfaces a visible error state (e.g. "Couldn't reach the server, please retry") instead of silently swallowing the failure.

## Severity

Low — noted in the plan as "low priority now that deployment is stable but should be fixed before next public launch."

nazarli-shabnam:
Already fixed in the current code — verified by reading the exact lines this issue cites.

Both call sites have a \`.catch()\` today:
- \`apps/ui/app/login/page.tsx\`: \`.catch(() => { /* API unreachable — stay on login page; user will see the form */ })\`
- \`apps/ui/app/setup/page.tsx\`: \`.catch(() => { /* API unreachable — stay on setup page; submission will show the error */ })\`

Neither surfaces a dedicated error banner, but the failure is no longer silently swallowed — the page falls back to rendering its form instead of hanging. Closing as resolved; feel free to reopen if a more visible error state is still wanted.

Issue #75, opened 2026-07-10
fix(ui): 401 mid-session doesn't clear auth state, causing a login redirect loop


nazarli-shabnam:
## Problem

When an API call returns 401, `handleResponse()` (`apps/ui/lib/api/client.ts:53-59`) removes the token from `localStorage` and dispatches a `clevis:unauthorized` event. `AuthGuard`'s handler for that event (`apps/ui/components/auth-guard.tsx:42-48`) only does `router.replace("/login")` — it never calls `logout()`, so `AuthProvider`'s React state (`user`, `token` in `apps/ui/lib/auth-context.tsx`) is never cleared.

## Failure scenario

A session expires (JWT expiry, or server-side revocation) while the user is active. The next API call 401s, the token is dropped from `localStorage`, but `useAuth().user` is still truthy in memory. The browser navigates to `/login`, whose effect (`apps/ui/app/login/page.tsx:43-47`) sees `user` truthy and immediately `router.replace(next)`, sending the user straight back to the protected page — which issues another request, gets another 401, and loops. From the user's perspective the app is stuck (either visibly bouncing between routes, or rendering as "logged in" while every request silently fails).

## Suggested fix

Have the `clevis:unauthorized` handler in `AuthGuard` call the existing `logout()` from `useAuth()` (which already clears `token`/`user` state and the localStorage key) before or instead of navigating, so `/login`'s "already authenticated" redirect can't fire on stale state.

## Severity

High — breaks the core login flow after any session expiry.

Tarekchehahde:
Addressed in batch PR #109 (local tests in commit message).

Tarekchehahde:
PR: https://github.com/nazarli-shabnam/clevis/pull/115

nazarli-shabnam:
Already fixed in the current code — verified by reading the exact lines this issue cites.

\`apps/ui/components/auth-guard.tsx\` (current \`handle401\`):
\`\`\`tsx
function handle401() {
  logout()
  router.replace("/login")
}
\`\`\`
It calls \`logout()\` before redirecting, so \`AuthProvider\`'s \`user\`/\`token\` state is cleared and \`/login\`'s "already authenticated" redirect can't fire on stale state. Closing.

PR #199, opened 2026-07-17
fix(ui,api): persist GitHub App installations via a post-install callback

### LINKED ISSUE #198: fix(ui,api): GitHub App installs are never persisted — no post-install callback or installation lookup
## Problem

After a user clicks \"Install GitHub App\" (`apps/ui/app/settings/page.tsx`, \`installUrl\` → `https://github.com/apps/{slug}/installations/new`), GitHub redirects back with `installation_id`/`setup_action` query params — but nothing in `apps/ui` reads them or calls the endpoints that would persist the installation.

`apps/api/src/routers/installations.py` confirms `POST /me/installations/sync` and `POST /orgs/{org}/installations/sync` are the *only* code paths that write `github_installations` rows (the webhook deliberately no-ops on `installation.created`, per its own comment, since it can't verify ownership). But:

- There is no page/route in `apps/ui` that handles GitHub's post-install redirect.
- `apps/ui/lib/api/client.ts` only defines `installations.list()` — no `sync` method exists client-side at all.
- Even if a callback page existed, the frontend has no way to resolve an `installation_id` to the account it belongs to (`account_login`/`account_type`) before calling `/sync`, since `UserOut` doesn't expose `github_login` and the sync endpoints require the caller to already supply those fields (`apps/api/src/schemas/installation.py`). No lookup endpoint existed.

## Failure scenario

A user signs in with GitHub, goes to Settings, clicks \"Install GitHub App\", picks an org (or their personal account) on GitHub's install screen, and is redirected back into the app. Nothing happens — no `github_installations` row is ever created. The Settings page's \"Connected organizations\" section still shows \"No organizations connected yet,\" and the sidebar profile widget (`apps/ui/components/app-sidebar.tsx:142-148`, itself only reading a `default_org` `localStorage` key regardless of real state) shows \"no org connected.\" There was previously no way to complete this connection through the UI at all.

## Fix

- Added `GET /me/installations/lookup/{installation_id}` (auth-required, read-only) that resolves an installation_id to its account via `github_app.get_installation` — reusing the error-mapping logic already in `_verify_installation`.
- Added `installations.lookup()` / `installations.sync()` to `apps/ui/lib/api/client.ts`.
- Added `apps/ui/app/settings/github-callback/page.tsx` — reads `installation_id`/`setup_action` from GitHub's redirect, looks up the account, and calls the matching `/me` or `/orgs/{org}` sync endpoint, with clear success/error/pending-approval states.
- Fixed `apps/ui/components/app-sidebar.tsx` to derive the displayed org from real `/me/orgs` membership data instead of only trusting the `default_org` localStorage key.

Note: this exact path (`/settings/github-callback`) must also be set as the GitHub App's \"Setup URL\" in the App's own settings on github.com — a one-time manual step outside this repo.

## Severity

High — this is the core "connect your GitHub org" flow; without it, installing the GitHub App had no effect.

nazarli-shabnam:
## Summary

Closes #198.

- Investigated a user report that "Sign in with GitHub" leads to a page saying "no org connected." Login/register themselves turned out to be fully functional — the real gap is that installing the GitHub App never persisted a `github_installations` row, because there was no UI callback to handle GitHub's post-install redirect, and no way to resolve an `installation_id` to its account before calling the existing sync endpoints.
- Added `GET /me/installations/lookup/{installation_id}` (`apps/api/src/routers/installations.py`) — auth-required, read-only, resolves an installation to `{account_login, account_type}` via the existing `github_app.get_installation`. Extracted the GitHub-call error-mapping shared with `_verify_installation` into `_fetch_installation` instead of duplicating it.
- Added `apps/ui/app/settings/github-callback/page.tsx` — reads `installation_id`/`setup_action` from GitHub's redirect, calls the new lookup endpoint, then calls the matching `/me` or `/orgs/{org}` sync endpoint, with explicit success / pending-approval / error states.
- Added `installations.lookup()` / `installations.sync()` to `apps/ui/lib/api/client.ts` (previously only `list()` existed — nothing client-side could call sync at all).
- Fixed `apps/ui/components/app-sidebar.tsx` to derive the displayed org from real `/me/orgs` membership data instead of trusting a `default_org` `localStorage` key unconditionally, which is what produced "no org connected" regardless of actual state.
- Documented the required GitHub App "Setup URL" config in `docs/self-hosting.md` — the callback page only gets hit if the App's own settings on github.com point at it; existing self-hosters need to add this manually (called out with a dedicated note for that case).
- While in the area, verified and closed 3 stale issues (#75, #96, #186) whose fixes were already present in the current code but never closed — comments cite the exact current lines proving each is resolved.

No database migration — `github_installations` schema already supported everything needed.

## Test plan
- [x] `pytest -q` — 242/242 passed (26 in `test_installations.py`, including 5 new tests for the lookup endpoint: success for User/Organization accounts, requires auth, 422 on nonexistent installation_id, 503 when the GitHub App isn't configured)
- [x] `bun run typecheck` — passes
- [x] `bun run lint` — passes (only pre-existing warnings in `coverage/`, unrelated)
- [x] `bun run build` — passes, `/settings/github-callback` compiles as a static route
- [x] `bun run test` (vitest) — passes
- [ ] Manual end-to-end walkthrough (sign in → Settings → Install GitHub App → redirect → row appears in `github_installations` → sidebar shows real org) requires a configured GitHub App (`GITHUB_APP_ID`/`GITHUB_APP_PRIVATE_KEY`/`NEXT_PUBLIC_GITHUB_APP_SLUG`) with its Setup URL pointed at this branch's deployed `/settings/github-callback` — not run in this environment; flagging for the reviewer to verify against a real GitHub App.


nazarli-shabnam (commented):
Summary of everything done:
- Guide: added the missing "Setup URL" step to docs/self-hosting.md's GitHub App registration section, including a callout for self-hosters who already registered their App before this change.
- PR #199 closes #198
- Test plan in the PR is honest about what's unverified: the actual end-to-end GitHub redirect can't be exercised in this environment without a configured GitHub App pointed at a deployed callback URL

PR #208, opened 2026-07-18
feat(api,ui): real GitHub Activity Feed (Phase 9, closes #176)

### LINKED ISSUE #176: feat(ui,api): Activity Feed only shows job queue, no GitHub event feed (Phase 9)
## Problem

`docs/plan.md`'s Phase 9 ("Activity Feed") is marked "⬜ planned." The existing `/activity` page (`apps/ui/app/activity/page.tsx`) is real and functional, but only surfaces the internal background-jobs queue — it is not a GitHub activity feed. There is no `GET /github/orgs/{org}/events` proxy endpoint anywhere in `apps/api/src/routers/`, and no event feed UI (pushes/PRs/issues/releases) with filter chips or auto-refresh.

## User impact

Users have no way to see what's actually happening across an org's repos (recent pushes, opened/merged PRs, new issues, releases) inside Clevis — "Activity" currently only means "job queue activity," which is a much narrower and less useful signal than the page name implies.

## Suggested fix

Add `GET /github/orgs/{org}/events` (proxying GitHub's org events API with server-side `summarize()` per `docs/plan.md`'s Phase 9 spec), and extend the Activity page with a real event feed: filter chips (`All`/`Pushes`/`PRs`/`Issues`/`Releases`), the existing job-queue panel kept as a side panel, and a 30s auto-refresh with a visible countdown. Note: `docs/plan.md` also flags this phase as eventually "re-pointed" to be event-sourced from webhooks once S3/S4 land — a polling-based version is still valid interim scope.

## Severity

Medium — the current `/activity` page's name doesn't match its content, and no GitHub-event visibility exists anywhere in the product.

nazarli-shabnam:
## Summary

Closes #176. `docs/plan.md`'s Phase 9 ("Activity Feed") was marked "planned" but never built — `/activity` only showed the internal job queue, not real GitHub org activity.

- **Backend:** `POST /github/orgs/{org_login}/events` (`apps/api/src/routers/github.py`) proxies GitHub's `/orgs/{org}/events`, filters out `[bot]` actors, and computes a human-readable `summary` server-side per event type (`PushEvent`, `PullRequestEvent`, `IssuesEvent`, `ReleaseEvent`, `CreateEvent`, with a raw-type fallback) — the frontend stays "dumb," no event-type switch logic in TypeScript.
- **Frontend:** the Activity page (`apps/ui/app/activity/page.tsx`) is now a two-column layout — the new `EventFeed` component (filter chips: All/Pushes/PRs/Issues/Releases, 30s auto-refresh with a visible "refreshes in Ns" countdown) as the primary panel, with the existing job-queue panel kept unchanged as a side panel.

## Why these specific choices

- **POST, not the literal `GET /github/orgs/{org}/events` the plan doc sketches.** Every other GitHub-token-bearing endpoint in this codebase (`apps/api/src/routers/repos.py`'s `RepoListInput`, `RepoStatsInput`, etc.) sends an optional client-supplied PAT in the POST body, specifically so a token never rides in a URL/query string where it could land in server logs, browser history, or a proxy's access log. This endpoint follows that same convention rather than the doc's shorthand.
- **Org/token resolution copies the Overview page's pattern** (`default_org` from `localStorage` → resolve token → auto-fetch), not the Repos/Security pages' manual "Load" button — a live feed should load itself, not wait for a click.
- **No new countdown component existed anywhere in the UI** — grepped the whole `apps/ui` tree for `setInterval`/countdown patterns first; this is genuinely new code (a small `useCountdown` hook local to the activity page), not a refactor of something existing.

## Deferred (explicitly out of scope, not in issue #176's stated ask)

- The Overview page's "Recent Events" badge (`docs/plan.md` §6.3) — not mentioned in the issue's "Suggested fix."
- A "Load more" pagination control — issue only asks for the feed itself; a single 30-event page is returned.
- Webhook-sourced event data (`docs/plan.md`'s S3/S4 migration note) — issue #176 itself says "a polling-based version is still valid interim scope."

None of these are schema/migration changes, and no sensitive-surface (RBAC/auth/token-encryption) logic was modified beyond adding one more consumer of the existing `resolve_org_token`/`require_org_role` helpers already used by `repos.py`.

## Test plan

- [x] `pytest -q apps/api/tests/test_github_events.py` — 8 new tests: per-event-type `summarize()` correctness, bot-actor exclusion, `per_page` passthrough, no-token → 400, GitHub 4xx/5xx → 400/503, unknown org → 404, non-member → 403.
- [x] `pytest -q` (full suite) — 313 passed, no regressions.
- [x] `bun run check` (typecheck + lint + build) — clean.
- [x] `bun run test` (full suite) — 165/166 passed; the 1 failure (`layout.test.tsx`, a Next/font-loading test) is a pre-existing flake under full-suite load, confirmed by rerunning it in isolation (passes).
- [x] New UI tests: `event-feed.test.tsx` (renders events, filter chip narrows results, empty state) and `activity-page.test.tsx` (configure-prompt when no org set, feed + job side-panel both render, countdown chip present).
- [ ] Not manually verified in-browser against a live GitHub org (no running dev stack in this session) — the mocked test coverage above is what backs this PR's correctness claim for now.


nazarli-shabnam:
Ran \`/code-review high\` against this diff and issue #176 (not the PR description). 9 findings surfaced across correctness, conventions, and efficiency; all fixed in the follow-up commit:

- An empty-body 200 from GitHub's events API was silently rendered as "no events" instead of surfacing an error (GitHubClient's `{}` fallback vs. an expected list).
- The "refreshes in Ns" countdown stuck at 0 forever after any single failed poll instead of tracking the real 30s cadence.
- `github.router` was the only org-scoped router mounted with an explicit prefix, inconsistent with every sibling (`repos.py`, `analytics.py`, etc.) and a likely 404 footgun for future contributors.
- `_github_error` was duplicated verbatim between `repos.py` and the new `github.py` — now shared.
- PushEvent summaries undercounted large pushes (GitHub truncates the embedded commit list at 20; now uses the true `payload.size`).
- Token resolution excluded valid 1-2 character org logins.
- A non-null assertion on the resolved token could throw under a query-key transition race — now a clear thrown error instead.
- No caching on the events endpoint despite 30s client-side polling and an existing TTL-cache precedent in `repos.py` for the same class of data — added a 25s cache.
- The countdown's 1s tick lived in the page component, forcing a full-tree re-render every second — isolated into its own leaf component.

Full pytest suite (317) and `bun run check`/`bun run test` green after the fixes.

PR #209, opened 2026-07-18
feat(checks,api): Security Dashboard v2 checks/history (Phase 10)

### LINKED ISSUE #177: feat(checks,api): Security Dashboard v2 checks/history not built (Phase 10)
## Problem

`docs/plan.md`'s Phase 10 ("Security Dashboard v2") is marked "⬜ planned." Today `packages/checks/src/checks/runner.py:11` hard-codes exactly three checks (`OrgMFARequired`, `BranchProtectionEnabled`, `SecretScanningEnabled` — all in `packages/checks/src/checks/github_checks.py`). There is no `DependabotAlertsCheck`, `CodeScanningCheck`, or `DefaultBranchNoForcePushCheck`, no `scan_results` table/migration to persist historical scan runs, and no `GET /analytics/history/{owner}` endpoint.

## User impact

Security coverage is limited to 3 of the checks the plan calls for, and every scan is a one-off snapshot — there's no way to see a security score's trend over time, which is one of the dashboard's core planned value props ("10-scan trend chart").

## Suggested fix

Add the three new `Check` subclasses (following the existing pattern in `github_checks.py`, registered in `runner.py`'s hard-coded list), add a `scan_results` table via a new Alembic migration to persist each analytics-overview run, add `GET /analytics/history/{owner}`, and extend the Security page with a score-gauge, donut, and 10-scan trend chart plus All/Failed/By-Severity tabs.

## Severity

Medium — directly limits the accuracy and usefulness of the product's core "security score" feature.

nazarli-shabnam:
## Summary

Closes #177. Implements Phase 10 ("Security Dashboard v2"), which was previously entirely unbuilt:

- **3 new security checks** (`packages/checks/src/checks/github_checks.py`): `DependabotAlertsCheck` (severity/critical-high aggregation), `CodeScanningCheck` (open alert counts), `DefaultBranchNoForcePushCheck` (force-push protection on the default branch). All three are registered in `runner.py`'s hard-coded check list and reuse the shared per-org repo list the runner already fetches once, matching the existing checks' pattern exactly.
- **Persisted scan history**: a new `scan_results` table (see migration note below), a `scan_results_repo`, and `GET /orgs/{org}/analytics/history` + `GET /me/analytics/history` endpoints (mirroring the existing overview endpoint pair). A row is inserted after every successful scan.
- **Security page redesign**: score gauge + donut now sit side-by-side, a new "Score trend (last N scans)" panel (reusing the existing `AreaTimeChart`) fed by the new history endpoint and refetched immediately after each scan, and the old status/severity `<select>` filters are replaced with `[All][Failed][By Severity]` tabs (styled like the Activity page's existing filter-chip pattern). `CheckValueDisplay` gains a `severity_counts` variant rendering color-coded chips for the Dependabot check's `{critical, high, medium, low}` shape.

## Why these specific decisions

- **Each new check treats a 404/403 from its per-repo endpoint as "not enabled for this repo," not an error** — Dependabot/code scanning being off for a given repo is expected and shouldn't fail the whole org-wide check. This mirrors `BranchProtectionEnabled`'s existing unknown/unprotected handling.
- **The `scan_results` insert lives in the router layer, not inside `analytics_service.py`** as the plan doc's literal wording suggests. `analytics_service.py` doesn't take a `db` param today and stays GitHub-only; routers already own persistence elsewhere in this codebase (e.g. `jobs.py`), so the insert happens in `org_analytics_overview`/`personal_analytics_overview` right after `_run_overview` succeeds. This is a deliberate deviation from the plan doc, not an oversight.
- **The code-scanning and force-push checks' values are rendered via the existing `ratio` display** (via new shape-sniffing branches in `normalizeCheckValue`), inverted so the numerator is always "clean repos out of total" — consistent with the existing `protected`/`enabled` ratio semantics. Only Dependabot's 4-bucket severity counts needed a new value type.
- **Org-scoped history endpoint exists server-side but isn't wired into the UI client** — same as the existing org-scoped overview endpoint, which the client also doesn't call. The personal `/me/...` path is what the Security page actually uses today.

## Schema change

**This PR includes a new Alembic migration**: `apps/api/alembic/versions/0015_add_scan_results.py`. It's additive-only — a new `scan_results` table (`id, owner, score, total_checks, failed_checks, checks_json, created_at`) plus a composite index on `(owner, created_at)`. No existing table, column, or constraint is altered or dropped. This was necessary because scan results previously had zero persistence — every scan was a one-off snapshot, which is exactly the gap issue #177 calls out ("no way to see a security score's trend over time"). Hand-authored (not autogenerated) to match `0001_initial_schema.py`'s existing `create_table` style; verified locally via `alembic upgrade head` against the dev Postgres container.

## Explicitly out of scope

- Any GitHub check beyond the 3 named in the issue (e.g. a dedicated secret-scanning *alerts* check, vs. the existing enablement-only `SecretScanningEnabled` — not requested).
- Wiring the org-scoped history endpoint into the UI client (parity endpoint only, matching the existing overview convention).
- Any FK relationship between `scan_results` and `orgs`/repos — kept as a loose string `owner` column, matching `SavedToken.org`'s existing style.

## Test plan

- [x] `pytest -q packages/checks/tests apps/api/tests` — 321 passed
- [x] `alembic upgrade head` against dev Postgres — migration applies cleanly
- [x] `bun run test` — 172 passed (27 files)
- [x] New backend tests cover: pass/fail/not_applicable per new check, 404-as-zero handling per check, history endpoint ordering/limit/scoping (org membership required, personal requires auth), insert-after-scan integration
- [x] New frontend tests cover: severity chips render/hide zero buckets, trend chart appears once 2+ history points exist, Failed tab narrows the check list
- [ ] Not manually verified in a running browser — no local dev server was started for this change; typecheck/lint/build/tests are the verification basis

## Update: code-review fixes (second commit)

`/code-review high` against issue #177's text surfaced two real bugs, fixed in a follow-up commit on this branch:

- **Authorization gap**: `GET /me/analytics/history` had no ownership check — any authenticated user could read any org's persisted scan history by naming it. Fixed with a new `scan_results.scanned_by_user_id` column and a membership/installation/prior-scan gate.
- **False "pass" under insufficient token scope**: `DependabotAlertsCheck`/`CodeScanningCheck` treated 403 the same as 404, silently reporting "clean" when the token just couldn't see the data. Now returns `error` when every repo was forbidden.

**This PR now includes a second schema change**: migration `0016_add_scan_results_scanned_by_user_id.py` — additive-only (new nullable FK column), no backfill or data-loss risk, needed to fix the authorization gap above (personal scans against an owner with no workspace Org row had no other way to record who's allowed to read that owner's history back).

See the review summary comment on this PR for full details.

### DIFF
--- a/apps/api/alembic/versions/0015_add_scan_results.py
+++ b/apps/api/alembic/versions/0015_add_scan_results.py
@@ -32,7 +32,7 @@ def upgrade() -> None:
         sa.Column("total_checks", sa.Integer(), nullable=False),
         sa.Column("failed_checks", sa.Integer(), nullable=False),
         sa.Column("checks_json", sa.Text(), nullable=False),
-        sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now()),
+        sa.Column("created_at", sa.DateTime(timezone=True), nullable=False, server_default=sa.func.now()),
     )
     op.create_index("ix_scan_results_owner_created_at", "scan_results", ["owner", "created_at"])
 
--- a/apps/api/alembic/versions/0015_add_scan_results.py
+++ b/apps/api/alembic/versions/0015_add_scan_results.py
@@ -1,0 +1,42 @@
+"""add scan_results
+
+Every /analytics/overview scan was previously a one-off snapshot with no
+history — there was no way to see a security score's trend over time. This
+adds a scan_results table that persists a row for every completed scan
+(owner, score, check counts, and the full checks payload as JSON text,
+matching the json-as-Text convention used by audit_logs.payload and
+jobs.payload/result elsewhere in this schema), plus a composite index on
+(owner, created_at) since every read of this table is "most recent N scans
+for this owner".
+
+Revision ID: 0015
+Revises: 0014
+Create Date: 2026-07-18
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "0015"
+down_revision = "0014"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "scan_results",
+        sa.Column("id", sa.Integer(), primary_key=True, autoincrement=True),
+        sa.Column("owner", sa.String(), nullable=False),
+        sa.Column("score", sa.Integer(), nullable=False),
+        sa.Column("total_checks", sa.Integer(), nullable=False),
+        sa.Column("failed_checks", sa.Integer(), nullable=False),
+        sa.Column("checks_json", sa.Text(), nullable=False),
+        sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now()),
+    )
+    op.create_index("ix_scan_results_owner_created_at", "scan_results", ["owner", "created_at"])
+
+
+def downgrade() -> None:
+    op.drop_index("ix_scan_results_owner_created_at", table_name="scan_results")
+    op.drop_table("scan_results")
--- a/apps/api/alembic/versions/0016_add_scan_results_scanned_by_user_id.py
+++ b/apps/api/alembic/versions/0016_add_scan_results_scanned_by_user_id.py
@@ -1,0 +1,40 @@
+"""add scan_results.scanned_by_user_id
+
+The new GET .../analytics/history endpoints read scan_results by a raw
+`owner` string with no other scoping. For the org-scoped endpoint that's
+fine (require_org_role already gates on membership), but the personal
+endpoint (GET /me/analytics/history?owner=...) had no such gate at all --
+any authenticated user could read any owner's persisted score history
+just by naming it, regardless of whether they have any access to that
+owner's GitHub data. A personal scan against an owner with no matching
+workspace Org (the common raw-PAT-paste flow, never installing the
+GitHub App) has no membership row to check against, so there was no way
+to scope personal-history reads without recording who ran the scan.
+
+This column tracks that: set on insert for personal scans only (org-scoped
+scans leave it null and rely on org membership at read time instead).
+Nullable and additive-only -- no backfill, no data-loss risk; existing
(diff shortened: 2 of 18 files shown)


nazarli-shabnam:
## /code-review high results

Reviewed the actual diff against issue #177's text (not this PR's description). Found and fixed:

1. **Authorization gap (fixed)** — `GET /me/analytics/history` had no ownership check beyond `require_auth`: any authenticated user could read any org's persisted scan-history summary just by naming the owner in the query string. Added `scan_results.scanned_by_user_id` (migration `0016`, additive-only) and a `_user_can_read_history` gate: allowed via org membership, a personal GitHub App installation for that owner, or having personally scanned that owner before. Denied (403) otherwise.
2. **False "pass" under insufficient token scope (fixed)** — `DependabotAlertsCheck`/`CodeScanningCheck` treated a 403 (permission denied) the same as a 404 (feature genuinely disabled), so a token missing the security-events scope silently reported a clean "pass" instead of surfacing that it couldn't see any data. Now returns `status="error"` when every repo in the org was forbidden, matching `BranchProtectionEnabled`'s existing unknown/error pattern.
3. **Duplicated insert block (fixed)** — factored the identical post-scan `scan_results_repo.insert(...)` call out of both overview handlers into a shared `_persist_scan` helper.
4. **Doubled branch-detail API calls (left unfixed, follow-up)** — `DefaultBranchNoForcePushCheck` re-fetches the exact same per-repo branch-detail endpoint `BranchProtectionEnabled` already calls, doubling that endpoint's GitHub API usage per scan. Sharing the fetch would require extending the `Check.run()` base signature across all six check classes — a larger architectural change than warranted in this pass. Flagging as a follow-up rather than rushing a coupling-heavy fix.

All 326 backend tests pass (including new tests for the auth gate and the 403-vs-404 distinction), migration `0016` applies cleanly, and the full suite from the original PR description still holds.

PR #211, opened 2026-07-18
feat(api,ui): read-only GitHub org roster for Collaborators page (Phase 11)

### LINKED ISSUE #178: feat(api): Collaborators page missing outside-collaborator/2FA data (Phase 11)
## Problem

`docs/plan.md`'s Phase 11 ("Collaborators") is marked "⬜ planned." Separately from the dead-stub-routing issue (#138), the page it should route to (`/settings/org/[login]/members`) only implements invitation create/list/revoke (`apps/api/src/routers/invitations.py`) — there is no `apps/api/src/routers/collab.py`, and none of `GET /github/orgs/{org}/members`, `.../outside_collaborators`, `.../invitations` (GitHub-side, distinct from Clevis's own invitation system), or `.../members/{username}/membership` exist.

## User impact

Even after #138 fixes the navigation dead-end, users still can't see the actual org roster (who has access, at what level), outside collaborators, or per-member 2FA status — only Clevis-internal pending invitations. This is the bulk of what a "Collaborators" page is supposed to show per the plan.

## Suggested fix

Add `apps/api/src/routers/collab.py` with the four GitHub-proxy endpoints above, and extend the Collaborators/Members UI with Members / Outside / Pending tabs and 2FA-status chips per member, per `docs/plan.md` Phase 11's spec.

## Severity

Medium — depends on #138 being resolved first to even be reachable, but is the larger of the two gaps.

nazarli-shabnam:
## Why

`docs/plan.md`'s Phase 11 ("Collaborators") was marked planned but unimplemented — the Collaborators/Members settings page only ever showed Clevis's own invite-by-email state (`apps/api/src/routers/invitations.py`). There was no way to see the actual GitHub org roster: who has access and at what level, outside collaborators, or per-member 2FA status. Closes #178.

## What changed

**Backend** — new `apps/api/src/routers/collab.py`, four read-only GitHub-proxy endpoints, all `GET`, all gated by `require_org_role(min_role="member")`, all resolving the org's token server-side via `resolve_org_token` (no client-supplied PAT override — these never carry a request body):

- `GET /github/orgs/{org_login}/members?role=all|member|admin` — GitHub's member list has no `role` field, so role is derived by diffing against a separate `role=admin` call. Also folds in a best-effort per-member 2FA overlay (`GET /orgs/{org}/members?filter=2fa_disabled`) rather than adding a 5th endpoint, since the UI always needs it joined against the member list. A 403/404 on that overlay call (token lacks org-owner scope) degrades gracefully — `two_factor_overlay_available: false`, each member's `two_factor_enabled: null` — rather than failing the whole request; any other error still propagates.
- `GET /github/orgs/{org_login}/outside_collaborators` — GitHub's endpoint returns *who*, not *which repos*, so this cross-references against `GET /orgs/{org}/repos`, capped at the first 50 repos (`repos_scanned`/`repos_total` in the response so the UI can show "scanned 50 of 140" rather than silently under-reporting on large orgs).
- `GET /github/orgs/{org_login}/invitations` — GitHub-side pending org invitations (distinct from Clevis's own invite-by-email system). Path is `/github/orgs/...` so there's no route collision with the existing `/orgs/{org_login}/invitations`.
- `GET /github/orgs/{org_login}/members/{username}/membership` — single-member state/role lookup.

No mutation endpoints — invite/revoke actions are explicitly out of scope per the plan doc (a later phase). No schema migration — nothing here is persisted; it's all a live GitHub read or a derived shape.

**Frontend** — extended `apps/ui/app/settings/org/[login]/members/page.tsx` in place: the existing Clevis invite-form/invitations table is kept as-is but relabeled "Clevis workspace invitations" to disambiguate it from a new "GitHub organization roster" section below it, with Members / Outside / Pending tabs (hand-rolled per the existing tab pattern in `apps/ui/app/security/page.tsx` — no Tabs primitive exists in this codebase). Each tab's query is gated with `enabled` so switching tabs doesn't fire every upstream GitHub call at once. Members show avatar/role badge/2FA chip (omitted, not shown as "unknown", when the overlay is unavailable) plus a "Members without 2FA: N" footer; Outside shows repo chips per collaborator plus a scan-cap note when relevant; Pending is explicitly labeled "Pending GitHub invitations" to avoid confusion with the Clevis section above it.

## Test plan
- [x] `pytest -q apps/api/tests/test_collab.py` — 26 new tests (role derivation, 2FA overlay success/degradation/non-degrading-on-other-errors, repo-scan cap, email-only invitations, RBAC, GitHub error mapping) — all pass
- [x] `pytest -q` (full backend suite) — 370 passed
- [x] `bun run typecheck && bun run lint` — clean
- [x] `bun run test` — 189/190 pass; the one failure (`layout.test.tsx`'s font-config test) is pre-existing and unrelated — a jsdom "navigation to another Document" limitation, not touched by this change
- [x] `bun run build` — production build succeeds

### DIFF
--- a/apps/api/src/main.py
+++ b/apps/api/src/main.py
@@ -11,6 +11,7 @@ from src.routers import (
     analytics,
     audit,
     auth,
+    collab,
     config,
     github,
     github_auth,
@@ -66,6 +67,7 @@ app.include_router(analytics.router, tags=["analytics"])
 app.include_router(actions_cache.router, tags=["actions-cache"])
 app.include_router(repos.router, tags=["repos"])
 app.include_router(github.router, tags=["github"])
+app.include_router(collab.router, tags=["collab"])
 app.include_router(jobs.router, prefix="/jobs", tags=["jobs"])
 app.include_router(audit.router, prefix="/audit", tags=["audit"])
 app.include_router(tokens.router, prefix="/tokens", tags=["tokens"])
--- a/apps/api/src/routers/collab.py
+++ b/apps/api/src/routers/collab.py
@@ -5,13 +5,19 @@ endpoints. No mutation routes here — invite/revoke actions are a later phase.
 
 Mounted with the full "/github/orgs/{org_login}/..." path on the router itself
 (no `prefix=` passed to include_router in main.py), matching github.py's
-convention for GitHub-proxy routers. Always resolves the org's installation
-token server-side via resolve_org_token (no client-supplied PAT override) —
-these are plain GETs, so there's nowhere to safely carry a body-level token.
+convention for GitHub-proxy routers. Resolves a token via resolve_org_token,
+preferring a GitHub App installation but falling back to a client-supplied PAT
+carried in the `X-GitHub-Token` request header -- these are plain GETs, so a
+PAT can't travel in the body the way every other GitHub-proxy POST route does;
+a header (never a query string, which would leak into logs/browser history) is
+the safe equivalent, matching the "PAT is optional if the App is connected"
+pattern used everywhere else in this codebase.
 """
 
+from concurrent.futures import ThreadPoolExecutor
+
 import httpx
-from fastapi import APIRouter, Depends, HTTPException
+from fastapi import APIRouter, Depends, Header, HTTPException
 from sqlalchemy.orm import Session
 from typing import Literal
 
@@ -37,9 +43,9 @@ router = APIRouter()
 _MAX_REPOS_SCANNED = 50
 
 
-def _resolve_token(db: Session, ctx: OrgContext, org_login: str) -> str:
+def _resolve_token(db: Session, ctx: OrgContext, org_login: str, client_token: str | None) -> str:
     try:
-        return resolve_org_token(db, org_id=ctx.org.id, account_login=org_login, client_token=None)
+        return resolve_org_token(db, org_id=ctx.org.id, account_login=org_login, client_token=client_token)
     except NoGitHubTokenAvailable as exc:
         raise HTTPException(status_code=400, detail=str(exc))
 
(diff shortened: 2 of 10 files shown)


nazarli-shabnam:
Ran an internal high-effort code review on this PR and pushed a follow-up commit addressing the findings:

- **PAT fallback gap (the significant one)**: the original 4 endpoints only ever resolved a GitHub App installation token — orgs without the App installed (relying on a pasted PAT, which every other GitHub-proxy feature in this app supports) got a hard 400 from the entire Collaborators roster. Fixed by accepting an optional client-supplied token via an `X-GitHub-Token` request header (not a query string, to avoid leaking it into logs/browser history) — these are GETs so it can't travel in a body like the POST-based GitHub-proxy routes do. The frontend now resolves a saved org token (same pattern as the Security page) and sends it through automatically.
- 2FA overlay now degrades gracefully on *any* GitHub error (not just 403/404 scope errors) — a transient network blip on that optional call no longer fails a response whose member/role data already succeeded.
- Parallelized the per-repo collaborator fan-out in `list_outside_collaborators` with a thread pool (matching `repos.py`'s existing pattern) instead of one sequential HTTP call per repo.
- Fixed a bare-dict-index KeyError risk on a GitHub invitation missing `created_at` — now skipped defensively instead of 500ing the whole list.
- Validated the roster tab id read from the URL query string so a stale/edited link falls back to the Members tab instead of rendering with no tab visually active.

29 backend tests / 373 full backend suite / typecheck+lint+7 relevant UI tests all pass after the fixes.

nazarli-shabnam (commented):
Coverage gaps are resolved.  Typecheck, lint, and build to be safe.

PR #213, opened 2026-07-22
feat(api,ui): Automation page workflow listing, run history, and dispatch (Phase 13)

### LINKED ISSUE #180: feat(api,ui): Automation page is a bare stub, no workflow dispatch (Phase 13)
## Problem

`docs/plan.md`'s Phase 13 ("Automation") is marked "⬜ planned." `apps/ui/app/automation/page.tsx` still renders a bare `EmptyState` with "Automation coming soon" (lines 14-17) and no query. There is no backend router for workflows at all — no `GET /github/repos/{owner}/{repo}/workflows`, `GET .../actions/runs`, or `POST .../workflows/{workflow_id}/dispatch`.

## User impact

The one existing "privileged action" pattern in the product (Actions-cache clear, with dry-run + audit trail) has no counterpart for triggering CI/CD workflows, despite the README describing "Automation (privileged jobs w/ dry-run + audit trail)" as a headline feature.

## Suggested fix

Add the workflow-listing/run-history/dispatch endpoints (dispatch gated behind the current auth model's admin-equivalent check — the plan's original spec references a `require_role("admin")` dependency that predates Phase 5's RBAC removal and needs updating to the current `require_org_role("admin")` pattern used elsewhere in `apps/api/src/routers/actions_cache.py`), and build the Automation page's workflow list, trigger modal, and run-duration chart. Per `docs/plan.md`'s S6 note, workflow dispatch is expected to remain a direct-GitHub-call action endpoint even after the aggregates migration lands, so this doesn't need to wait on S2-S6.

## Severity

Medium — entirely unbuilt end-to-end (UI and API), and referenced in the README as a current feature.

nazarli-shabnam:
## Summary

Closes #180. Implements docs/plan.md Phase 13 ("Automation") end-to-end: workflow listing, run history, and workflow dispatch, both backend and frontend. Previously the Automation page was a bare `EmptyState` stub with no backend router at all.

## What changed and why

**Backend** — new `apps/api/src/routers/automation.py`, mounted in `main.py` alongside the other GitHub-proxy routers:
- `GET /orgs/{org_login}/repos/{owner}/{repo}/workflows` and the `/me/repos/{owner}/{repo}/workflows` personal-scoped equivalent — lists a repo's workflows. Overlays each workflow's most-recent-run status/conclusion/timestamp via a second, best-effort GitHub call (degrades gracefully like `collab.py`'s 2FA overlay — a failure here doesn't fail the whole response).
- `GET .../actions/runs` (+ `/me/...` equivalent) — run history with a computed `duration_ms` (from `run_started_at`/`updated_at`, `None` while a run is still in progress).
- `POST .../workflows/{workflow_id}/dispatch` (+ `/me/...` equivalent) — the one write endpoint, gated behind `require_org_role(min_role="admin")` for the org-scoped route. The issue's suggested fix flagged that the plan's original `require_role("admin")` spec predates Phase 5's RBAC removal — this uses the current `require_org_role("admin")` pattern from `actions_cache.py` instead. Writes an `audit_logs` row **before** the GitHub call (same order as `actions_cache.py`'s dry-run+audit pattern), so there's a record even if GitHub rejects the dispatch. There's no `dry_run` mode here — GitHub Actions has no dispatch-preview API, so the destructive-action UI reuses the arm/confirm two-click pattern from `CachePanel` but drops the separate "dry run" button that pattern also has.
- Reads use `X-GitHub-Token` (header, never query string) matching `collab.py`'s convention since these are GETs; the write uses a body token field matching `actions_cache.py`.
- New `apps/api/tests/test_automation.py` (12 tests): RBAC (member/admin/outsider), token-resolution fallback, overlay-failure degradation, duration computation, and — importantly — that the audit log is written even when the subsequent GitHub call fails.

**Frontend** — replaces the `apps/ui/app/automation/page.tsx` stub:
- Owner/repo/token config panel (same shape as `ReposPage`/`SecurityPage`), a "Load workflows" action that fires both the workflows and runs queries together.
- Workflow table with last-run status icons, a "Dispatch" action per row that opens an inline ref input + arm/confirm dispatch button (reusing `CachePanel`'s destructive-action UX, since no Dialog primitive exists in this repo yet).
- Run-duration bar chart (`BarGroupChart`) and a run-history table.
- New `api.automation.*` namespace in `lib/api/client.ts`, new types in `lib/api/types.ts`.
- New `apps/ui/tests/components/automation-page.test.tsx` (4 tests): initial empty state, load flow, arm/confirm dispatch (asserts the API is only called after the *second* click), and error surfacing.

No schema migration — this is all live GitHub reads plus one `audit_logs` write, no new persisted state. No `.env`/RBAC-config changes; the auth model itself (`require_org_role`) is unchanged, just applied to a new route.

## Test plan

- [x] `pytest -q apps/api/tests/test_automation.py` — 12/12 pass
- [x] `pytest -q` (full backend suite) — 406/406 pass
- [x] `bun run test` (full UI suite) — 215/215 pass

### DIFF
--- a/apps/api/src/main.py
+++ b/apps/api/src/main.py
@@ -11,6 +11,7 @@ from src.routers import (
     analytics,
     audit,
     auth,
+    automation,
     collab,
     config,
     github,
@@ -68,6 +69,7 @@ app.include_router(actions_cache.router, tags=["actions-cache"])
 app.include_router(repos.router, tags=["repos"])
 app.include_router(github.router, tags=["github"])
 app.include_router(collab.router, tags=["collab"])
+app.include_router(automation.router, tags=["automation"])
 app.include_router(jobs.router, prefix="/jobs", tags=["jobs"])
 app.include_router(audit.router, prefix="/audit", tags=["audit"])
 app.include_router(tokens.router, prefix="/tokens", tags=["tokens"])
--- a/apps/api/src/routers/automation.py
+++ b/apps/api/src/routers/automation.py
@@ -1,0 +1,237 @@
+"""Workflow listing/run-history/dispatch endpoints (docs/plan.md Phase 13 — Automation).
+
+Reads are plain GETs with an optional client-supplied PAT carried in the
+`X-GitHub-Token` header (never a query string), matching collab.py's convention.
+Dispatch is a write, so it stays a POST with the token in the body, matching
+actions_cache.py's convention -- and per docs/plan.md's cross-cutting note,
+dispatch is expected to remain a direct-GitHub-call action endpoint even after
+the aggregates migration lands, so it doesn't wait on that work.
+
+Dispatch is gated behind org-admin (require_org_role(min_role="admin")) --
+GitHub Actions has no dispatch-preview API, so unlike actions-cache clear there
+is no dry-run mode here. The audit log is written before the GitHub call so
+there's a record even if GitHub rejects or times out the dispatch.
+"""
+
+from datetime import datetime
+
+import httpx
+from fastapi import APIRouter, Depends, Header, HTTPException
+from sqlalchemy.orm import Session
+
+from src.core.auth import UserOut, require_auth
+from src.core.db import get_db
+from src.core.rbac import OrgContext, assert_owner_matches_org, require_org_role
+from src.repositories import audit_repo
+from src.schemas.automation import (
+    DispatchInput,
+    DispatchResponse,
+    RunSummary,
+    RunsResponse,
+    WorkflowSummary,
+    WorkflowsResponse,
+)
+from src.services.github_client import GitHubClient, github_error as _github_error
+from src.services.token_resolution import NoGitHubTokenAvailable, resolve_org_token, resolve_personal_token
+
+router = APIRouter()
+
+
+def _run_duration_ms(run: dict) -> int | None:
+    started = run.get("run_started_at")
+    updated = run.get("updated_at")
+    if not started or not updated or run.get("status") != "completed":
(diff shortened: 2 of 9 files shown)


nazarli-shabnam:
### Code review

No issues found at the confidence bar for this pass (highest-scoring candidate: 75/100, below the 80 threshold). Checked for bugs, AGENTS.md compliance, and consistency with the sibling patterns this PR deliberately mirrors (`actions_cache.py`, `collab.py`, `cache-panel.tsx`).

One near-threshold finding was applied anyway since it was cheap and clearly correct: the token-resolve gate on the new page used `owner.trim().length > 2`, which was already identified and fixed as a bug in `activity/page.tsx` (commit 6589d49 — valid GitHub org logins can be 1-2 characters). Fixed in [d0e531c](https://github.com/nazarli-shabnam/clevis/commit/d0e531c). Note: `repos/page.tsx` and `security/page.tsx` still carry the same unfixed `> 2` gate — out of scope for this PR, worth a follow-up.

nazarli-shabnam (commented):
Reviewed against #180 — this is solid. RBAC on dispatch correctly uses `require_org_role("admin")` (not the stale header-based check), the audit log write is genuinely ordered before the GitHub call and survives a failed dispatch, and the arm/confirm UI is a real two-click flow. Thanks for the clean work here.

One small non-blocking follow-up for later: the workflow list and last-run overlay aren't paginated, so repos with a lot of workflows/runs could show a truncated view — worth a follow-up issue, not a blocker for this PR.

PR #228, opened 2026-07-22
feat(api,ui): security compliance matrix and secret-scanning alerts (Phase 16)

### LINKED ISSUE #182: feat(api,ui): Security compliance matrix and secret-scanning alerts missing (Phase 16)
## Problem

`docs/plan.md`'s Phase 16 ("Security: Expanded Coverage") is marked "⬜ planned." There is no `GET /analytics/security-matrix/{owner}` (per-repo × dimension compliance matrix) or `GET /github/repos/{owner}/{repo}/secret-scanning` endpoint, and no compliance-matrix table, stacked vulnerability-by-severity chart, secret-scanning alerts list, or remediation trend chart in the UI.

## User impact

Security posture today is only visible as an aggregate org-wide score (see Phase 10 gap) — there's no per-repo breakdown showing which specific repos are failing which specific checks, which is what an admin actually needs to act on findings.

## Suggested fix

Add the two endpoints above and the four UI elements per `docs/plan.md` Phase 16's spec. Note the plan explicitly calls out that the secret-scanning alerts list must never display the actual secret value, only alert metadata.

## Severity

Low-Medium — depends on Phase 10's expanded checks landing first for full value, but the compliance-matrix view is independently useful with the existing 3 checks.

nazarli-shabnam:
## Summary

Closes #182. Implements docs/plan.md Phase 16 ("Security: Expanded Coverage"): a per-repo compliance-matrix view and secret-scanning alert metadata, both backend and frontend.

## What changed and why

**Backend** — new `apps/api/src/routers/security.py` + `apps/api/src/schemas/security.py`:

- `GET /me/analytics/security-matrix/{owner}` — for each repo (capped at 20, since this costs up to 3 GitHub calls per repo: branch details, Dependabot alerts, code-scanning alerts), computes `branch_protection`, `secret_scanning` (read straight off the repo object's `security_and_analysis` field, no extra call), `dependabot_enabled` + critical/high counts, `code_scanning` (clear/not), `force_push_allowed`, and a 0–100 `score` across those 5 dimensions. Any dimension that can't be resolved (network error, no visibility) degrades to a non-penalizing default rather than 500ing the response, matching the fan-out pattern already used in `analytics.py`. A `MatrixSummary` aggregates fully-compliant count, critical-risk count, secret-hit count, and severity totals.
- `GET /me/repos/{owner}/{repo}/secret-scanning` — alert list, **metadata only**: `number, state, secret_type, secret_type_display, resolved_reason, created_at, resolved_at, repo, url`. The actual secret value is never included — enforced by the Pydantic model's field list (there's no field to populate it into), with an explicit `# NOTE` comment, and a dedicated test asserting the raw secret string never appears anywhere in the response body even when GitHub's raw payload happens to include one.
- **Scope note on RBAC**: I originally wrote both endpoints as org-scoped (`require_org_role`), mirroring `collab.py`/`actions_cache.py`. Partway through I realized that's the wrong fit here — the existing Security page scans an arbitrary GitHub org by name via `/me/analytics/overview`, with no workspace-`Org`/`OrgMembership` requirement. An org-role-gated matrix endpoint would 404/403 for exactly the case the page already supports (an owner the caller has a token for but no workspace Org record for). Switched both to personal-scoped (`/me/...`, `require_auth` + `resolve_personal_token`) to match. Noting this since it's a real RBAC-adjacent design decision, not a mechanical one.
- These re-derive each dimension directly from GitHub per repo rather than reusing `checks.github_checks`' check classes, since those aggregate across all repos into one org-wide pass/fail — not built to return a per-repo row.
- 14 new backend tests.

**Frontend** — `apps/ui/app/security/page.tsx`:
- "Run scan" now also fires the compliance-matrix fetch alongside the existing org-wide scan.
- Compliance Matrix table (per-repo, clickable rows to select a repo for the alerts panel).
- Vulnerabilities-by-repo chart (`BarGroupChart`, critical/high stacked per repo).
- Secret Scanning Alerts panel with a repo selector and an explicit "Secret values are never shown here — metadata only" note next to the list, per the issue's requirement.
- Remediation Trend chart — reuses the existing scan-history data (`total_checks - failed_checks` per historical scan) rather than the literal per-severity Dependabot trend docs/plan.md sketches, since that would need the API to expose each historical scan's full `checks_json` breakdown (not currently returned by `GET /me/analytics/history`). Commented inline as a scoped approximation, not silently substituted.
- New `api.security.matrix` / `api.security.secretScanning` client methods, new types.
- 2 new tests in `security-page.test.tsx` (matrix load + repo-row rendering, and an explicit "no raw secret value ever renders" assertion), plus existing tests updated with the new mocks.

No schema migration. No `.env`/secrets changes. RBAC: see the scope note above — this PR intentionally does NOT add a `require_org_role` gate, for the reason described.

## Test plan

- [x] `pytest -q` (full backend suite) — 402/402 pass
- [x] `bun run test` (full UI suite) — 213/213 pass (one `layout.test.tsx` timeout during a concurrent run was confirmed flaky/contention-only — passes in isolation)

### DIFF
--- a/apps/api/src/main.py
+++ b/apps/api/src/main.py
@@ -21,6 +21,7 @@ from src.routers import (
     jobs,
     orgs,
     repos,
+    security,
     tokens,
     webhooks,
 )
@@ -68,6 +69,7 @@ app.include_router(actions_cache.router, tags=["actions-cache"])
 app.include_router(repos.router, tags=["repos"])
 app.include_router(github.router, tags=["github"])
 app.include_router(collab.router, tags=["collab"])
+app.include_router(security.router, tags=["security"])
 app.include_router(jobs.router, prefix="/jobs", tags=["jobs"])
 app.include_router(audit.router, prefix="/audit", tags=["audit"])
 app.include_router(tokens.router, prefix="/tokens", tags=["tokens"])
--- a/apps/api/src/routers/security.py
+++ b/apps/api/src/routers/security.py
@@ -1,0 +1,189 @@
+"""Security compliance matrix and secret-scanning alerts (docs/plan.md Phase 16).
+
+The org-wide `analytics.overview` scan already computes org-level pass/fail check
+results (see `checks.runner.run_all_checks`) -- this router instead breaks the same
+dimensions down per-repo, which is what an admin needs to act on a specific finding.
+It re-derives each dimension directly from the GitHub API per repo rather than reusing
+`checks.github_checks`, since those check classes are written to aggregate across all
+repos into a single org-wide pass/fail, not to return a per-repo row.
+
+Personal-scoped (`/me/...`), matching `analytics.py`'s `/me/analytics/overview` --
+the Security page scans an arbitrary owner by name, not a workspace Org the caller
+necessarily has an `OrgMembership` row for, so `require_org_role` doesn't apply here.
+"""
+
+from concurrent.futures import ThreadPoolExecutor
+
+import httpx
+from fastapi import APIRouter, Depends, Header, HTTPException
+from sqlalchemy.orm import Session
+
+from src.core.auth import UserOut, require_auth
+from src.core.db import get_db
+from src.schemas.security import (
+    MatrixSummary,
+    RepoSecurityRow,
+    SecretAlert,
+    SecretScanningResponse,
+    SecurityMatrixResponse,
+    VulnCounts,
+)
+from src.services.github_client import GitHubClient, github_error as _github_error
+from src.services.token_resolution import NoGitHubTokenAvailable, resolve_personal_token
+
+router = APIRouter()
+
+# Each repo costs up to 3 additional GitHub calls (branch, dependabot, code-scanning)
+# on top of the initial repo list, so the cap here is tighter than the single-call
+# aggregate helpers in analytics.py.
+_MAX_REPOS_FOR_MATRIX = 20
+
+
+def _repo_row(client: GitHubClient, owner: str, repo: dict) -> RepoSecurityRow:
+    name = repo["name"]
+    branch = repo.get("default_branch")
+
+    branch_protection = False
+    force_push_allowed = False
+    try:
+        details = client.request("GET", f"/repos/{owner}/{name}/branches/{branch}")
+        if isinstance(details, dict):
+            branch_protection = bool(details.get("protected"))
+            protection = details.get("protection") or {}
+            force_push_allowed = bool((protection.get("allow_force_pushes") or {}).get("enabled"))
+    except (httpx.HTTPStatusError, httpx.RequestError):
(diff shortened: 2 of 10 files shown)


nazarli-shabnam:
### Code review

Found 1 issue, confirmed independently by four review passes (compliance audit, bug scan, git-history, and comment-guidance checks all converged on it):

1. Per-repo dimensions that couldn't actually be evaluated (403/429/network error on branch-protection, Dependabot, or code-scanning calls) were silently scored as compliant instead of excluded — the same false-pass class of bug already fixed once at the org-aggregate level for `DependabotAlertsCheck`/`CodeScanningCheck` (see git history around commit `3184c76`). A repo the token can't see into was scoring identically to, or better than, a repo genuinely clean.

https://github.com/nazarli-shabnam/clevis/blob/32e388b7a99cf2afa5aca1aa973c75c3c1cc971f/apps/api/src/routers/security.py#L61-L100

Fixed in [470844a](https://github.com/nazarli-shabnam/clevis/commit/470844ae1a1d64c34e45b0180ff590c66d98a6fa): added `unknown_dimensions` to `RepoSecurityRow`, excluded from the score denominator, and distinguished a genuine 404 ("this is off") from an inconclusive error/403/429, matching the existing `packages/checks/src/checks/github_checks.py` convention. Also fixed an adjacent bug found during the same pass: `secret_type_display` was reading a GitHub field name that doesn't exist (`secret_type_display` vs. the real `secret_type_display_name`), so it silently fell back to the raw type slug on every alert.

nazarli-shabnam (commented):
reviewed against https://github.com/nazarli-shabnam/clevis/issues/182. The core security requirement holds: the secret-scanning response schema has no field capable of carrying a raw secret value, and the test actually injects a fake secret into the raw payload and asserts it never appears in the response, so that's genuinely verified, not just asserted.

However, this isn't merge-ready yet: when both the security scan and the new compliance-matrix panel fail (e.g. no GitHub App installation for the org), the page renders two visually-identical, unlabeled error banners with the exact same text — the existing scan-error banner and the new matrix panel's error text both surface the same resolve_personal_token error string verbatim. That's a real UX bug (a user can't tell these are two different failures), and it's also the direct cause of the E2E failure at e2e/security-settings.spec.ts:93 (strict-mode violation — the text locator now matches 2 elements).

Requested fix: reword the compliance-matrix panel's error text at apps/ui/app/security/page.tsx:384 to be distinguishable from the scan-error banner (e.g. prefix it with "Compliance matrix unavailable: ..."), and firm up the e2e assertion at security-settings.spec.ts:93 to target the specific banner rather than a bare page-wide text match, so a future duplicate-text panel doesn't reintroduce this. Pushing a fix commit for this now.

PR #232, opened 2026-07-23
fix(ui): thread invite/next context through GitHub OAuth sign-in

### LINKED ISSUE #222: fix(ui): GitHub OAuth sign-in drops invite context, unlike password login
## Problem

`apps/ui/app/invite/[token]/page.tsx` correctly encodes `next=/invite/{token}` for the password-login link (line 55). But the GitHub OAuth button on `apps/ui/app/login/page.tsx:162` always does `window.location.href = ${API_BASE}/auth/github/login` unconditionally, ignoring any `next` param entirely.

## Failure scenario

A user clicks an invite link, lands on the invite preview page, and chooses "Sign in with GitHub" instead of password login. After the OAuth round-trip, they land on the dashboard with no memory of the invite — they have to manually remember and re-navigate to the original invite URL to accept it, or the invite is effectively abandoned.

## Suggested fix

Thread `next`/the invite token through the GitHub OAuth login URL (e.g. as a query param preserved through `/auth/github/login` → `/auth/github/callback` → final redirect), the same way the password-login path already does.

## Severity

Medium — a real dead-end in the invite acceptance flow for anyone whose primary auth method is GitHub OAuth (likely the majority of users, given this is a GitHub analytics tool).

nazarli-shabnam:
## What changed

Fixes #222. The password-login path (`apps/ui/app/invite/[token]/page.tsx:55`) already carried `next=/invite/{token}` through `/login`, so a password sign-in resumes the invite afterward. The "Sign in with GitHub" button on `apps/ui/app/login/page.tsx` ignored `next` entirely, always sending the browser to `${API_BASE}/auth/github/login` with no param — so anyone who chose GitHub sign-in from an invite link landed on the dashboard with no memory of the invite.

**Frontend (`apps/ui/app/login/page.tsx`):** the GitHub button now appends `?next=<path>` to the OAuth login URL whenever `next` isn't the default `/`. Also removed the "GitHub sign-in returns you to the dashboard instead" caveat from `apps/ui/app/invite/[token]/page.tsx` now that it's no longer true.

**Backend — this is the sensitive part of the change** (touches OAuth/session code, flagging per AGENTS.md):
- `apps/api/src/services/github_oauth.py`: `sign_state()` now accepts an optional `next` param and embeds it in the signed OAuth-state JWT (the same token already used for CSRF protection between `/login` and `/callback`) rather than introducing a new cookie or query param to carry it through the GitHub round-trip. Added `safe_next_path()` — a server-side port of the frontend's `safeNextPath()` allowlist (same-origin relative paths only, rejects `//`, backslashes, control chars) — applied both when accepting the incoming `next` query param and again when decoding it back out of the state (`decode_state_next()`), so a malformed or external value can never end up in a redirect.
- `apps/api/src/routers/github_auth.py`: `/login` accepts `?next=`, passes it into `sign_state()`. `/callback` decodes `next` from the verified state and redirects there instead of always going to `_ui_redirect_target()` (the site root).
- No new state, no new cookie, no change to the existing CSRF nonce-cookie binding — `next` just rides along in the payload of the state token that was already being signed and verified.

## Test plan

- [x] `apps/api/tests/test_github_oauth.py` — new tests for `safe_next_path` (accepts/rejects), `sign_state`/`decode_state_next` round-trip with and without a valid/unsafe `next`.
- [x] `apps/api/tests/test_github_auth.py` — new router tests: `/login?next=` embeds it in the state; unsafe `next` is dropped at `/login`; `/callback` redirects to the next path on a full successful flow; falls back to root with no `next`; falls back to root when an unsafe `next` was requested. Also added the `_reset_rate_limiter` autouse fixture (mirroring the existing one in `test_auth.py`) — the module-level in-memory rate-limit bucket on `/auth/github/callback` is shared across tests in the same process, and the new tests' extra callback calls pushed the file over the default 10-request window, causing an unrelated pre-existing test to flake with a 429. Reset it the same way `test_auth.py` already does for `/auth/login`.
- [x] `pytest -q apps/api apps/worker` — 443 passed.

### DIFF
--- a/apps/api/src/routers/github_auth.py
+++ b/apps/api/src/routers/github_auth.py
@@ -89,9 +89,9 @@ def _clear_state_cookie(response: Response) -> None:
 
 
 @router.get("/login")
-def github_login(request: Request):
+def github_login(request: Request, next: str | None = None):
     try:
-        state, nonce = github_oauth.sign_state()
+        state, nonce = github_oauth.sign_state(next=next)
         url = github_oauth.build_authorize_url(state=state, redirect_uri=_callback_url(request))
     except github_oauth.GitHubOAuthNotConfigured:
         return _ui_login_error_redirect("github_not_configured")
@@ -124,6 +124,7 @@ def github_callback(
         error_response = _ui_login_error_redirect("github_invalid_state")
         _clear_state_cookie(error_response)
         return error_response
+    next_path = github_oauth.decode_state_next(state)
     try:
         user_token = github_oauth.exchange_code_for_token(code, redirect_uri=_callback_url(request))
         identity = github_oauth.fetch_identity(user_token)
@@ -144,7 +145,8 @@ def github_callback(
         return error_response
     org_provisioning.sync_org_admin_memberships(db, user, user_token)
     token = create_access_token(user.id, user.email, user.is_workspace_admin, user.name, user.token_version)
-    response = RedirectResponse(_ui_redirect_target(), status_code=303)
+    redirect_target = f"{_ui_redirect_target()}{next_path}" if next_path else _ui_redirect_target()
+    response = RedirectResponse(redirect_target, status_code=303)
     set_session_cookie(response, token)
     _clear_state_cookie(response)
     return response
--- a/apps/api/src/services/github_oauth.py
+++ b/apps/api/src/services/github_oauth.py
@@ -69,12 +69,31 @@ def _web_base() -> str:
     return base
 
 
-def sign_state(*, now: int | None = None) -> tuple[str, str]:
+def safe_next_path(next_path: str | None) -> str | None:
+    """Only accept unambiguous same-site relative paths -- mirrors safeNextPath() in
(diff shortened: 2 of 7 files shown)


nazarli-shabnam:
### Code review

No issues found. Checked for bugs (including open-redirect/CSRF interaction, since this touches the OAuth login/callback flow), CLAUDE.md/AGENTS.md compliance, and prior review history on these files.

Note: a prior PR (#65) had a coderabbitai review flagging CWE-601 open-redirect via a backslash-smuggled `next` value in the frontend's `safeNextPath()`. That was fixed then, and this PR's new backend `safe_next_path()` in [`apps/api/src/services/github_oauth.py`](https://github.com/nazarli-shabnam/clevis/blob/cefdf38bf742af613beec278f086213c99fbffbd/apps/api/src/services/github_oauth.py#L72-L83) correctly mirrors the same hardened validation (rejects `//`, backslashes, control chars), applied both when signing and when decoding the state.

PR #234, opened 2026-07-23
fix(checks): cap unbounded GitHub API pagination in _get_all_pages

### LINKED ISSUE #214: fix(checks): unbounded GitHub API pagination in _get_all_pages can spike memory to OOM
## Problem

`packages/checks/src/checks/github_checks.py:_get_all_pages()` (lines 38-56) follows GitHub's `Link: rel=\"next\"` header in an uncapped `while url:` loop, calling `results.extend(r.json())` on every page with no limit on page count or total result size:

```python
def _get_all_pages(base_url: str, path: str, token: str) -> list:
    ...
    results = []
    url: str | None = f"{base_url}{path}?per_page=100"
    with httpx.Client(timeout=20) as client:
        while url:
            r = _get_with_retry(client, url, headers)
            r.raise_for_status()
            results.extend(r.json())
            url = None
            for part in r.headers.get("Link", "").split(","):
                ...
    return results
```

This runs inside the `api` process during every analytics/security-check scan (`analytics_service.py` → `checks.runner.run_all_checks`), and again for the initial repo-list prefetch in `packages/checks/src/checks/runner.py:17`.

## Failure scenario

Run a scan against a GitHub org with a large number of repos, org members, or workflow runs (thousands+). Each page (100 items) is fully materialized and appended to one in-memory list with no cap. For a large enough org, a single scan can pull tens of thousands of objects into memory in one process — this is a plausible direct cause of the repeated OOM-driven crashes seen after deploys, since a fresh scan is often triggered right after a redeploy.

## Suggested fix

Add a hard cap on pages/total results (fail gracefully or truncate with a warning past N pages), and/or stream-process each page (aggregate incrementally) instead of accumulating the full list before returning.

## Severity

High — directly implicated in production OOM crashes requiring manual redeploys, not just a correctness issue.

nazarli-shabnam:
## What changed

Fixes #214. `_get_all_pages()` in `packages/checks/src/checks/github_checks.py` followed GitHub's `Link: rel="next"` header in an uncapped `while url:` loop, calling `results.extend(r.json())` on every page with no limit on page count or total result size.

## Why

This runs inside the `api` process on every analytics/security-check scan (`analytics_service.py` → `checks.runner.run_all_checks`), and again for `runner.py`'s org repo-list prefetch. Against a large enough org (thousands of repos, org members, or workflow runs), a single scan can pull tens of thousands of objects into one in-memory list with no cap — a plausible direct cause of the OOM-driven crashes reported after deploys, since a fresh scan is often triggered right after a redeploy.

## Fix

- Added a `_MAX_PAGES = 50` module-level constant (50 pages × `per_page=100` = 5,000 items cap), following this file's existing pattern of hardcoded module constants (retry count, timeout, `per_page` are all inline literals already — no env/config precedent in this package to break).
- Once the cap is hit, the loop stops fetching further pages and **logs a warning instead of raising**. `runner.py`'s prefetch call (`repos = _get_all_pages(...)` at line 17) isn't wrapped in try/except, so a raised exception there would crash the entire `run_all_checks()` call rather than degrading gracefully — truncating with a warning matches the issue's own suggested fix ("fail gracefully ... or truncate with a warning") and keeps a scan usable (partial results) instead of losing it entirely.
- All 5 call sites in this file (`BranchProtectionEnabled`, `SecretScanningEnabled`, `DependabotAlertsCheck`, `CodeScanningCheck`, `DefaultBranchNoForcePushCheck`) and the `runner.py` prefetch are pure iteration/`len()` consumers of the returned list — no other logic needed to change.

## Test plan

- [x] `packages/checks/tests/test_github_checks.py` — new tests: `_get_all_pages` still correctly follows the `Link` header across pages under the cap; and truncates at exactly `_MAX_PAGES` pages with a warning logged when more pages are available, following the existing `httpx.Client.get` monkeypatch/mock pattern already used in this file's retry tests.
- [x] `pytest -q packages/checks apps/api apps/worker` — 464 passed.

### DIFF
--- a/packages/checks/src/checks/github_checks.py
+++ b/packages/checks/src/checks/github_checks.py
@@ -1,8 +1,19 @@
+import logging
 import time
 
 import httpx
 
 from checks.base import Check, CheckMetadata
+
+logger = logging.getLogger(__name__)
+
+# Hard cap on pages fetched by _get_all_pages -- at per_page=100 this is 5,000 items max.
+# Without this, a large enough org (thousands of repos/members/workflow runs) can pull an
+# unbounded number of objects into one in-memory list during a scan, which is a plausible
+# direct cause of OOM crashes in the api process (see issue #214). Truncating with a warning
+# degrades a scan's completeness rather than crashing the whole run_all_checks() call, since
+# the runner.py prefetch that calls this isn't wrapped in try/except.
+_MAX_PAGES = 50
 
 
 def _get_with_retry(client: httpx.Client, url: str, headers: dict) -> httpx.Response:
@@ -43,16 +54,26 @@ def _get_all_pages(base_url: str, path: str, token: str) -> list:
     }
     results = []
     url: str | None = f"{base_url}{path}?per_page=100"
+    pages_fetched = 0
     with httpx.Client(timeout=20) as client:
         while url:
             r = _get_with_retry(client, url, headers)
             r.raise_for_status()
             results.extend(r.json())
+            pages_fetched += 1
             url = None
             for part in r.headers.get("Link", "").split(","):
                 part = part.strip()
                 if 'rel="next"' in part:
                     url = part.split(";")[0].strip().strip("<>")
+            if url and pages_fetched >= _MAX_PAGES:
+                logger.warning(
+                    "Truncating pagination for %s after %d pages (%d items) -- more pages were available",
+                    path,
+                    pages_fetched,
+                    len(results),
+                )
+                break
     return results
 
 
--- a/packages/checks/tests/test_github_checks.py
+++ b/packages/checks/tests/test_github_checks.py
@@ -6,6 +6,7 @@ import httpx
 import pytest
 
 from checks.github_checks import (
+    _MAX_PAGES,
     BranchProtectionEnabled,
     CodeScanningCheck,
     DefaultBranchNoForcePushCheck,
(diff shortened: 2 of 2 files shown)


nazarli-shabnam:
### Code review

No blocking issues found. Verified the page-counting logic (stops at exactly `_MAX_PAGES`, no off-by-one, no infinite-loop risk) and confirmed the commit message's claim that all 5 checks consuming `_get_all_pages`'s return value are plain `len()`/iteration consumers.

Two non-blocking observations for awareness, neither in scope for this PR:

1. [`apps/api/src/services/github_client.py`'s `request_paginated`](https://github.com/nazarli-shabnam/clevis/blob/bc59d7968ef197298740a2bd4ade20a9039eeb18/apps/api/src/services/github_client.py#L45-L74) has the same unbounded `while url:` pagination pattern and isn't capped by this PR. Issue #214 specifically named `_get_all_pages` in `packages/checks`, so this is a candidate follow-up issue rather than something this PR should absorb.
2. A truncated scan now silently reports as a normal "pass"/"fail" result (e.g. `SecretScanningEnabled` compliant = enabled == total, computed only over the truncated `repos` list) with no signal beyond the new WARNING log line. This is the exact tradeoff issue #214 itself asked for ("truncate with a warning ... instead of crashing"), so not a defect in this PR, just worth knowing if a future change wants to surface "partial scan" status further up the stack (e.g. in the analytics report shown to users).

PR #235, opened 2026-07-23
fix(infra): add container memory limits and worker/ui healthchecks

### LINKED ISSUE #216: fix(infra): no container memory limits or worker/ui healthchecks in docker-compose
## Problem

`docker-compose.yml` sets no `mem_limit` / `deploy.resources.limits` on any service (`db`, `api`, `worker`, `ui`), and only `db` and `api` have a `healthcheck` block — `worker` and `ui` have none, only `restart: unless-stopped`.

## Failure scenario

On a memory-constrained host, any one service creeping up in RSS (see #214 — unbounded pagination in `packages/checks` is one concrete trigger, running inside the `api` process) has nothing stopping it from consuming host memory until the kernel OOM-killer intervenes. Without per-container limits, the OOM killer's victim selection isn't necessarily the actual offender, so the failure can present as an unrelated service crashing — consistent with the reported symptom of "the app crashes on almost every deploy" with no clear single cause and needing a manual redeploy. Separately, because `worker` and `ui` have no healthcheck, a hung-but-technically-alive container (not fully crashed) won't be auto-restarted by Docker's restart policy — only a hard crash does.

## Suggested fix

- Add `mem_limit` (or `deploy.resources.limits.memory` if using swarm-style config) to each service, sized to the host's actual RAM budget, so one service can't starve the others.
- Add `healthcheck` blocks to `worker` and `ui` (worker: e.g. a liveness file/heartbeat touch; ui: an HTTP check against a Next.js route) so hangs trigger `restart: unless-stopped` the same way crashes do.

## Severity

High — this is very likely the direct cause of the recurring "crashes on deploy, needs manual redeploy" behavior described by the maintainer.

nazarli-shabnam:
## What changed — this touches deploy-relevant infra config (`docker-compose.yml`), flagging per AGENTS.md

Fixes #216. `docker-compose.yml` set no `mem_limit`/`deploy.resources.limits` on any service, and only `db`/`api` had a `healthcheck` — `worker` and `ui` had none, only `restart: unless-stopped`.

## Why

On a memory-constrained host, one service creeping up in RSS (e.g. the unbounded-pagination OOM risk fixed in #214, which runs inside the `api` process) had nothing stopping it from consuming host memory until the kernel OOM-killer intervened. Without per-container limits, the OOM killer's victim selection isn't necessarily the actual offender, so the failure can present as an unrelated service crashing — consistent with the reported "crashes on almost every deploy" with no clear single cause. Separately, without a healthcheck, a hung-but-technically-alive `worker`/`ui` container wouldn't be auto-restarted — only a hard crash triggers `restart: unless-stopped`.

## Fix

- **Memory limits** (asked the maintainer for the target host's specs, given a 4GB host budget): `db: 1g`, `api: 768m`, `worker: 512m`, `ui: 512m` — leaves ~1.25g headroom for the OS/Traefik. Uses `mem_limit` (the classic-compose key), not `deploy.resources.limits.memory` — this file has no top-level `version:`/swarm mode, so `deploy.*` would be silently ignored under plain `docker compose up`. Comment in the file notes these are tuned for a 4GB host and should be revisited if that changes.
- **Worker healthcheck**: `apps/worker/src/worker.py` now touches a heartbeat file (`/tmp/worker_heartbeat`) once per poll-loop iteration. The healthcheck CMD is a `python` one-liner checking the file's mtime is <60s old, matching the `api` service's existing `python -c` style (avoids relying on curl/wget, which aren't guaranteed on `python:3.12-slim`).
- **UI healthcheck**: added `apps/ui/app/api/health/route.ts` — a plain 200 response, deliberately not calling the API or DB so it can't report "unhealthy" due to a downstream outage the UI container isn't responsible for. Healthcheck CMD uses `node`, not curl, since the `ui` image's runner stage (`node:22-alpine`) doesn't have curl by default.

## Test plan

- [x] `docker compose config` — parses cleanly, confirms `mem_limit` and both new `healthcheck` blocks apply correctly (didn't run a full `docker compose up --build`, which is slower and wasn't necessary to validate the compose syntax and app-level changes).
- [x] `apps/worker/tests/test_heartbeat.py` — new tests: `_touch_heartbeat` writes a current timestamp; doesn't raise if the path is unwritable (heartbeat failures must never take down job processing).
- [x] `apps/ui/tests/components/health-route.test.ts` — new test: `GET /api/health` returns 200 with no API/DB calls.
- [x] `pytest -q apps/worker apps/api packages/checks` — 464 passed.
- [x] `bun run test` — 257 passed, 1 pre-existing flaky failure in `layout.test.tsx` (a jsdom "navigation to another Document" timeout, unrelated to this change) — passes in isolation.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -79,7 +79,7 @@ Key variables:
 - `GITHUB_API_BASE` — default `https://api.github.com`; set for GitHub Enterprise (e.g. `https://github.yourco.com/api/v3`). Used by both the API and the worker. Not runtime-editable because it's where GitHub tokens are sent.
 
 **DB-backed config (editable in Settings → Instance Configuration):**
-- `worker_poll_seconds` — default `5`; the worker re-reads it each loop, so changes take effect live without a restart
+- `worker_poll_seconds` — default `5`, clamped to `[1, 30]`; the worker re-reads it each loop, so changes take effect live without a restart. The upper clamp keeps the worker's heartbeat healthcheck in `docker-compose.yml` meaningful (see `apps/worker/src/worker.py`'s `_MAX_POLL_SECONDS`).
 
 ## Running locally
 
--- a/apps/ui/app/api/health/route.ts
+++ b/apps/ui/app/api/health/route.ts
@@ -1,0 +1,6 @@
+// Liveness endpoint for the docker-compose healthcheck -- deliberately does not call the
+// API or DB, so it can't report "unhealthy" due to a downstream outage the UI container
+// itself isn't responsible for. It only confirms the Next.js server is up and responding.
+export function GET() {
+  return new Response("ok", { status: 200 })
+}
--- a/apps/ui/tests/components/health-route.test.ts
+++ b/apps/ui/tests/components/health-route.test.ts
@@ -1,0 +1,10 @@
+import { describe, expect, it } from "vitest";
+import { GET } from "@/app/api/health/route";
+
+describe("GET /api/health", () => {
+  it("returns 200 without calling the API or DB", async () => {
+    const res = GET();
+    expect(res.status).toBe(200);
+    expect(await res.text()).toBe("ok");
+  });
+});
--- a/apps/worker/src/worker.py
+++ b/apps/worker/src/worker.py
@@ -22,6 +22,14 @@ log = logging.getLogger(__name__)
 # from a genuinely healthy one -- `restart: unless-stopped` only fires on a hard crash,
 # not a hang, so without this a stuck worker container would never be restarted.
 HEARTBEAT_FILE = Path("/tmp/worker_heartbeat")
+# The docker-compose healthcheck treats the heartbeat file as stale past 60s (see
+# docker-compose.yml). worker_poll_seconds is a live-editable app_config value with no
+# upper bound otherwise, and the heartbeat only gets touched once per loop iteration --
+# an operator setting it above this cap would make every iteration's normal sleep alone
+# exceed the healthcheck's staleness threshold, permanently false-positive-ing the worker
+# as hung. Kept comfortably below 60s so real job processing inside an iteration still
+# has margin before the healthcheck's threshold is reached.
+_MAX_POLL_SECONDS = 30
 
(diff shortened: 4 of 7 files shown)


nazarli-shabnam:
### Code review

Found 1 issue, fixed in a follow-up commit on this branch:

1. The new worker healthcheck treats the heartbeat file as stale past 60s, but `worker_poll_seconds` is a live-editable `app_config` value with no upper bound (`_read_poll_seconds` only clamped to a minimum of 1), and the heartbeat is only touched once per poll-loop iteration. An operator setting `worker_poll_seconds` above ~55s — a supported, documented config value per AGENTS.md's "the worker re-reads it each loop, so changes take effect live without a restart" — would make every iteration's normal sleep alone exceed the healthcheck's 60s threshold, permanently marking a healthy worker as hung.

Fixed by clamping `worker_poll_seconds` to `[1, 30]` via a new `_MAX_POLL_SECONDS` in [`apps/worker/src/worker.py`](https://github.com/nazarli-shabnam/clevis/blob/60c0b006e47ca04cf774807449bb0da5ad0f4c22/apps/worker/src/worker.py#L25-L32), with tests in `apps/worker/tests/test_poll_seconds.py` and the `AGENTS.md` doc updated to note the new range.

Two non-blocking observations, out of scope for this PR:

- Under plain `docker compose up` (no swarm mode, no watchdog), a container's `restart: unless-stopped` policy only restarts on process exit, not on `HEALTHCHECK` reporting `unhealthy` — this applies equally to the pre-existing `db`/`api` healthchecks, not something this PR introduces. The new healthchecks do correctly surface hang detection (`docker ps` status) and gate `depends_on: condition: service_healthy` at startup, but won't by themselves auto-restart a hung worker/ui container later at runtime without an additional mechanism (e.g. a `docker-autoheal`-style sidecar).
- `apps/api/src/services/github_client.py`'s `request_paginated` has the same unbounded `while url:` pagination pattern that was just capped in `packages/checks/src/checks/github_checks.py` (issue #214 / PR #234) — already flagged on that PR's review thread as a candidate follow-up, not fixed here since issue #216 is about limits/healthchecks, not that pagination helper.

Issue #221, opened 2026-07-22
fix(ui): 'Members without 2FA' count ignores the active roster search filter


nazarli-shabnam:
## Problem

`apps/ui/app/settings/org/[login]/members/page.tsx` — the members table (line 74 area) renders `filteredMembers` (filtered by the search box), but the "Members without 2FA" footer count (line 178) is computed from the unfiltered `membersQuery.data.members`.

## Failure scenario

An admin types a search query into the roster search box. The table below shows a filtered subset of members, but the "Members without 2FA" count next to/below it still reflects the entire org roster — two numbers on screen that look related but describe different sets, which reads as a bug or a data-integrity issue to anyone reading the page.

## Suggested fix

Compute the 2FA count from `filteredMembers` (matching whatever's currently visible in the table), or clearly label it as an org-wide total independent of the active filter.

## Severity

Medium — misleading security-relevant number (2FA compliance) shown alongside a filtered view, likely to cause confusion during actual security review.

nazarli-shabnam:
Already fixed on main in 284e120 (`fix(ui): ui polish batch...`) — `membersWithout2fa` is now derived from `filteredMembers` rather than the unfiltered roster (apps/ui/app/settings/org/[login]/members/page.tsx). That commit was pushed directly rather than via a PR referencing this issue, so it never auto-closed. No further action needed.

PR #236, opened 2026-07-23
fix(ui): UI polish batch (collaborators, roster filters, repo-segment charset, security retry)

### LINKED ISSUE #226: fix(ui): minor UI polish backlog (collaborators empty state, roster filter latency, repo-segment validation, security tab retry)
## Problem

Four small, independent UI rough edges found during a bug-hunt sweep — batched into one issue as none are independently severe enough to justify a solo issue, but each is a good scoped pickup.

**1. Collaborators page silent dead-end** — `apps/ui/app/collaborators/page.tsx:19-25`. If a user has GitHub org memberships but no `admin` role in any of them, `adminOrgs` is empty and the page shows a generic "No organization to manage" empty state — it never surfaces which orgs the user *is* a member of (just not admin), forcing them to dig through Settings to discover that. By-design but a rough edge worth smoothing.

**2. Roster search/role-filter latency mismatch** — `apps/ui/app/settings/org/[login]/members/page.tsx`. `roleFilter` changes trigger a full server refetch (it's part of the TanStack Query key, ~line 54) while `search` is purely client-side filtering. The two controls look like one filter bar but behave with different latency, and there's no loading indicator distinguishing "refetching by role" from "typing search" — a role change while a search term is present can flash an empty/mismatched table momentarily.

**3. `parseOwnerRepo` has no charset validation** — `apps/ui/lib/repo-segment.ts:5`. Splits on `~` and requires exactly 2 parts, which is currently safe since GitHub names can't contain `~`, but there's no validation that `owner`/`repo` match GitHub's actual allowed charset before being passed to `api.repos.stats(...)` — relies entirely on the backend to reject bogus values. Defense-in-depth gap, not currently exploitable.

**4. Security tab has no retry-on-error affordance** — `apps/ui/app/repos/[repo]/page.tsx`. Once `securityQuery.isError`, switching away and back to the Security tab shows the same cached error (TanStack Query default behavior) without retrying — the only recovery path from a transient failure is a full page reload.

## Suggested fix

- #1: list the user's non-admin org memberships in the empty state, with a note on how to request admin access.
- #2: add a distinct loading indicator for role-filter refetches vs. client-side search filtering.
- #3: validate `owner`/`repo` against GitHub's allowed username/repo charset before use.
- #4: add a manual "Retry" button/action on the Security tab's error state.

## Severity

Low for each — polish items, not correctness or security bugs.

nazarli-shabnam:
## What changed

Fixes #226 (4-item polish bundle) and #221 (bundled into the same PR since it's the same file as one of #226's items).

1. **Collaborators page dead-end** (`apps/ui/app/collaborators/page.tsx`) — when a user has org memberships but no `admin` role in any of them, the page now lists which orgs they're a member of instead of a bare "No organization to manage" message, so they don't have to dig through Settings to find out.
2. **"Members without 2FA" ignores active search filter (#221)** (`apps/ui/app/settings/org/[login]/members/page.tsx:100`) — now derives from `filteredMembers` instead of the unfiltered roster, so it can't disagree with what's visibly in the table.
3. **Roster search/role-filter latency mismatch** (same file) — added a distinct "Updating…" indicator for role-filter refetches (server-side, part of the TanStack Query key) vs. the client-side search filter. Also switched the query to `placeholderData: keepPreviousData` so a role change no longer flashes an empty table while refetching — this ended up being the more correct fix for the underlying "looks like one filter bar, behaves with different latency" complaint than just adding a spinner would have been.
4. **`parseOwnerRepo` has no charset validation** (`apps/ui/lib/repo-segment.ts`) — added GitHub's actual allowed owner/repo charset validation (previously only checked for a single `~` separator) as defense-in-depth before the value reaches `api.repos.stats(...)`.
5. **Security tab has no retry-on-error affordance** (`apps/ui/app/repos/[repo]/page.tsx`) — added a Retry action reusing the existing `SectionError` component, extracted from `apps/ui/app/settings/page.tsx` (where it was a local, unexported function) into `apps/ui/components/section-error.tsx` so both pages can share it, instead of duplicating it.

## Test plan

- [x] `apps/ui/tests/lib/repo-segment.test.ts` — new tests for valid/invalid charset segments.
- [x] `apps/ui/tests/components/collaborators-page.test.tsx` — updated for the new non-admin-memberships listing; added a case for the truly-empty-memberships generic message.
- [x] `apps/ui/tests/components/github-roster.test.tsx` — new tests: 2FA count follows search filter; "Updating…" shows during a role-filter refetch and clears after.
- [x] `apps/ui/tests/components/repo-detail-page.test.tsx` — new test: clicking Retry on a failed security fetch re-fetches and renders the result.
- [x] `bun run test` (touched files) — 58 passed.

### DIFF
--- a/apps/ui/app/collaborators/page.tsx
+++ b/apps/ui/app/collaborators/page.tsx
@@ -50,6 +50,21 @@ export default function CollaboratorsPage() {
             title="Couldn't load your organizations"
             description="Something went wrong checking your organization memberships. Try refreshing the page."
           />
+        ) : memberships.length > 0 ? (
+          <div className="border border-dashed border-border rounded-md px-6 py-12">
+            <p className="text-sm text-muted-foreground">
+              Member management is only available for organizations where you&rsquo;re an admin. You&rsquo;re a
+              member (not admin) of:
+            </p>
+            <ul className="mt-2 text-sm text-foreground list-disc list-inside">
+              {memberships.map((m) => (
+                <li key={m.org_login}>{m.org_login}</li>
+              ))}
+            </ul>
+            <p className="mt-3 text-xs text-muted-foreground">
+              Ask an admin of one of these organizations to grant you admin access.
+            </p>
+          </div>
         ) : (
           <EmptyState
             title="No organization to manage"
--- a/apps/ui/app/repos/[repo]/page.tsx
+++ b/apps/ui/app/repos/[repo]/page.tsx
@@ -9,6 +9,7 @@ import Link from "next/link"
 import { useMutation, useQuery } from "@tanstack/react-query"
 import { PageHeader } from "@/components/page-header"
 import { Skeleton } from "@/components/ui/skeleton"
+import { SectionError } from "@/components/section-error"
 import { GitPullRequest, ArrowSquareOut, Star, GitFork, Eye, ShieldCheck, ShieldWarning, Shield } from "@phosphor-icons/react"
 import { api } from "@/lib/api/client"
 import { parseOwnerRepo } from "@/lib/repo-segment"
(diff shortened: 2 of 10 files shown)


nazarli-shabnam:
### Code review

No issues found. Checked for bugs (including the new regex charset validation, the keepPreviousData/2FA-count interaction, and the SectionError extraction's prop wiring), CLAUDE.md/AGENTS.md compliance, and prior review history on these files.

Worth noting: the collaborators-page fix directly resolves a UX gap that was flagged as an unresolved open question in an earlier PR (#149) — non-admin org members previously had no fallback state.

PR #238, opened 2026-07-23
fix(api): batch of small low-severity gaps (case-sensitive login check, unbounded query params, unbounded free-text fields, unguarded pagination prefetch)

### LINKED ISSUE #224: fix(api): batch of small low-severity gaps (case-sensitive login check, unbounded query params, unbounded free-text fields, unguarded pagination prefetch)
## Problem

Four small, independent, low-severity gaps found during a bug-hunt sweep — batched into one issue since none warrant a solo issue on their own, but each is a good single-PR pickup for a new contributor.

**1. Case-sensitive GitHub login comparison** — `apps/api/src/core/rbac.py:48-53` (`assert_owner_matches_org`) does `owner != ctx.org.github_login`, a strict comparison. GitHub logins are case-insensitive (`Acme` and `acme` are the same account). Contrast with `apps/api/src/routers/installations.py:67`, which correctly does `.lower() != .lower()` for the same class of comparison. Fails closed (denies access) rather than being exploitable, but is a usability bug and an inconsistency with the rest of the codebase.

**2. Unbounded `per_page` on automation endpoints** — `apps/api/src/routers/automation.py:157` and `:211` declare `per_page: int = 10` with no `Query(le=..., ge=...)` bound, unlike `apps/api/src/routers/audit.py:27`'s equivalent `limit: int = Query(default=100, le=500)`. A caller can pass an arbitrary/negative value, producing an opaque upstream GitHub 400 instead of a clean 422 at the boundary.

**3. Unbounded free-text fields flowing into JSON payload columns** — `apps/api/src/schemas/cache.py:10-14` (`CacheClearInput.key`/`.ref`) and `apps/api/src/schemas/automation.py:41-46` (`DispatchInput.ref`/`.inputs`) have no `max_length` constraint, flowing into `jobs.payload` / `audit_logs.payload` (both unbounded `Text` columns). A caller can submit an arbitrarily large value, bloating those tables — a cheap DoS vector via legitimate authenticated endpoints.

**4. Repo-list prefetch not defended like individual checks** — `packages/checks/src/checks/runner.py:17`, the initial `_get_all_pages(...)` call for the org's repo list sits outside the `try/except Exception` block that wraps each individual `check.run(...)` call (lines 29-36). Every check is hardened to degrade to `{"status": "error", ...}` on failure, but if this initial fetch itself raises, the whole `run_all_checks` function raises instead of returning a partial/error result — an inconsistency in the function's own error-handling contract (currently not exploitable through the one existing caller, which happens to catch the relevant exception types upstream, but latent for any future caller).

## Suggested fix

- #1: change to `.lower()` comparison, matching `installations.py`.
- #2: add `Query(ge=1, le=100)` (or similar) bounds matching the `audit.py` pattern.
- #3: add reasonable `max_length` constraints to the affected Pydantic fields.
- #4: wrap the repo-list fetch in the same try/except pattern used for individual checks, returning an error shape consistent with the rest of the function.

## Severity

Low for each — none independently exploitable/severe, but all are quick, well-scoped fixes.

nazarli-shabnam:
## What changed

Fixes #224 — four small, independent, low-severity gaps found during a bug-hunt sweep, bundled into one PR since none warranted a solo PR on their own.

1. **Case-sensitive GitHub login comparison** (`apps/api/src/core/rbac.py:52`, `assert_owner_matches_org`) — did `owner != ctx.org.github_login`, a strict comparison. GitHub logins are case-insensitive (`Acme` and `acme` are the same account). Contrast with `apps/api/src/routers/installations.py:67`, which correctly does `.lower() != .lower()` for the same class of comparison. Fails closed (denies access) rather than being exploitable, but was a usability bug and an inconsistency. Now matches `installations.py`'s pattern.
2. **Unbounded `per_page` on automation endpoints** (`apps/api/src/routers/automation.py:157,211`) — declared `per_page: int = 10` with no `Query(le=..., ge=...)` bound, unlike `apps/api/src/routers/audit.py:27`'s equivalent `limit: int = Query(default=100, le=500)`. A caller could pass an arbitrary/negative value, producing an opaque upstream GitHub 400 instead of a clean 422 at the boundary. Added `Query(ge=1, le=100)`.
3. **Unbounded free-text fields flowing into JSON payload columns** (`apps/api/src/schemas/cache.py:10-14`'s `CacheClearInput.key`/`.ref` and `apps/api/src/schemas/automation.py:41-46`'s `DispatchInput.ref`/`.inputs`) — no `max_length` constraint, flowing into `jobs.payload`/`audit_logs.payload` (both unbounded `Text` columns). A caller could submit an arbitrarily large value, bloating those tables — a cheap DoS vector via legitimate authenticated endpoints. Added `max_length` constraints: 512 for cache `key` (matches GitHub's own Actions cache-key limit), 255 for `ref` fields (generous git-ref-name cap), and for dispatch `inputs` — max 10 keys, 1024 chars per value (matching GitHub's own `workflow_dispatch` API limits).
4. **Repo-list prefetch not defended like individual checks** (`packages/checks/src/checks/runner.py:17`) — the initial `_get_all_pages(...)` call for the org's repo list sat outside the `try/except Exception` block that wraps each individual `check.run(...)` call. Every check is hardened to degrade to `{"status": "error", ...}` on failure, but if this initial fetch itself raised, the whole `run_all_checks` function raised instead of returning a partial/error result. Wrapped it in the same pattern — on failure, every check now gets the same `"status": "error"` shape instead of the exception propagating uncaught.

## Test plan

- [x] `apps/api/tests/test_rbac.py` — new unit tests for `assert_owner_matches_org`: exact match, case-insensitive match, rejects a different org.
- [x] `apps/api/tests/test_automation.py` — new tests: `per_page` out of `[1,100]` returns 422; oversized dispatch `ref`/too many `inputs` return 422.
- [x] `apps/api/tests/test_actions_cache.py` — new test: oversized cache `key`/`ref` return 422.
- [x] `packages/checks/tests/test_runner_isolation.py` — new test: a repo-list prefetch failure degrades all 6 checks to `"error"` status instead of raising out of `run_all_checks`.
- [x] `pytest -q` — 479 passed.

### DIFF
--- a/apps/api/src/core/rbac.py
+++ b/apps/api/src/core/rbac.py
@@ -49,5 +49,5 @@ def assert_owner_matches_org(owner: str, ctx: OrgContext) -> None:
     """Raises 403 if a repo-level `owner` path/body value doesn't match the org context
     require_org_role already resolved — keeps an org-scoped route from acting on a
     GitHub owner outside the org the caller was authorized for."""
-    if owner != ctx.org.github_login:
+    if owner.lower() != ctx.org.github_login.lower():
         raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="owner must match the org in the URL")
--- a/apps/api/src/routers/automation.py
+++ b/apps/api/src/routers/automation.py
@@ -16,7 +16,7 @@ there's a record even if GitHub rejects or times out the dispatch.
 from datetime import datetime
 
 import httpx
-from fastapi import APIRouter, Depends, Header, HTTPException
+from fastapi import APIRouter, Depends, Header, HTTPException, Query
 from sqlalchemy.orm import Session
 
 from src.core.auth import UserOut, require_auth
@@ -154,7 +154,7 @@ def org_list_runs(
     org_login: str,
     owner: str,
     repo: str,
-    per_page: int = 10,
+    per_page: int = Query(default=10, ge=1, le=100),
     ctx: OrgContext = Depends(require_org_role(min_role="member")),
     db: Session = Depends(get_db),
     x_github_token: str | None = Header(default=None),
@@ -208,7 +208,7 @@ def personal_list_workflows(
 def personal_list_runs(
     owner: str,
     repo: str,
-    per_page: int = 10,
+    per_page: int = Query(default=10, ge=1, le=100),
     user: UserOut = Depends(require_auth),
     db: Session = Depends(get_db),
     x_github_token: str | None = Header(default=None),
--- a/apps/api/src/schemas/automation.py
+++ b/apps/api/src/schemas/automation.py
@@ -45,7 +45,8 @@ class DispatchInput(BaseModel):
     token: SecretStr | None = None
     # Bounded so a caller can't bloat the jobs/audit_logs payload columns (both
     # unbounded Text) with an arbitrarily large value. 255 is a generous cap for a git
-    # ref name; GitHub's own workflow_dispatch API caps inputs at 10 keys, values 1024 chars.
+    # ref name. GitHub's workflow_dispatch API documents a hard limit of 10 input keys;
+    # 1024 chars per value isn't a documented GitHub limit, just a generous defensive cap.
     ref: str = Field(max_length=255)
     inputs: dict[str, Annotated[str, Field(max_length=1024)]] | None = Field(default=None, max_length=10)
 
--- a/apps/api/src/schemas/automation.py
+++ b/apps/api/src/schemas/automation.py
@@ -1,6 +1,7 @@
 from datetime import datetime
+from typing import Annotated
 
-from pydantic import BaseModel, SecretStr
+from pydantic import BaseModel, Field, SecretStr
 
 
 class WorkflowSummary(BaseModel):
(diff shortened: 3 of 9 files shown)


nazarli-shabnam:
### Code review

Found 1 issue, fixed in a follow-up commit on this branch:

1. The `DispatchInput.inputs` comment in `apps/api/src/schemas/automation.py` stated GitHub's `workflow_dispatch` API caps input values at "1024 chars" as if that were a documented GitHub limit, when only the 10-key cap is actually documented — 1024 chars was a self-chosen defensive bound presented as fact (AGENTS.md's "don't fabricate facts" guardrail).

Fixed in [`ebb45eb`](https://github.com/nazarli-shabnam/clevis/blob/ebb45eb1b3dd8b9fcfe9a8d426c9ef8b8451f067/apps/api/src/schemas/automation.py#L46-L50) by correcting the comment to distinguish the documented 10-key limit from the self-chosen 1024-char value cap.

Otherwise clean — verified all 4 fixes are correct and mutually non-interfering (case-insensitive comparison can't cause a false match, `per_page` bound doesn't break any existing caller, the `dict` `max_length` pattern is valid pydantic v2 and confirmed by its own regression test, and the `runner.py` failure path returns the same response shape `analytics_service.py` already handles). Git history confirms none of these were previously-deliberate behavior being regressed.

PR #239, opened 2026-07-23
fix(api,checks): GitHub secondary rate limit (403) is not retried like 429 is

### LINKED ISSUE #219: fix(api,checks): GitHub secondary rate limit (403) is not retried like 429 is
## Problem

`packages/checks/src/checks/github_checks.py:_get_with_retry()` (lines 8-23) and `apps/api/src/services/github_client.py:GitHubClient.request()`/`request_paginated()` (lines 27-74) both only back off/retry on `httpx.RequestError` or an HTTP 429 response. GitHub's secondary/abuse rate limit commonly returns **403** instead (often with a `Retry-After` header), especially under the concurrent fan-out this codebase already uses (`ThreadPoolExecutor` calls in `repos.py`, `collab.py`, `analytics.py`).

## Failure scenario

Run a scan or dispatch several concurrent org/repo API calls against a busy org close to GitHub's abuse-detection thresholds. GitHub returns 403 for a request that would otherwise succeed on retry. Because 403 isn't in the retry branch, `raise_for_status()` fires immediately and the whole check/analytics call surfaces as a hard failure (a whole security check reported as `"error"`, or a whole cockpit/analytics endpoint failing) instead of transparently retrying like the 429 case does.

## Suggested fix

Treat 403 responses that include rate-limit-indicating headers (`Retry-After`, or `X-RateLimit-Remaining: 0`) the same as 429 in both retry loops — back off and retry rather than raising immediately.

## Severity

Medium — surfaces as intermittent, hard-to-reproduce check/analytics failures under load, with a straightforward fix shared across two call sites that already have the retry contract in place.

nazarli-shabnam:
## What changed

Fixes #219. `packages/checks/src/checks/github_checks.py:_get_with_retry()` and `apps/api/src/services/github_client.py`'s `GitHubClient.request()`/`request_paginated()` only backed off/retried on `httpx.RequestError` or an HTTP 429. GitHub's secondary/abuse rate limit commonly returns **403** instead (often with a `Retry-After` header), especially under the concurrent fan-out this codebase already uses (`ThreadPoolExecutor` calls in `repos.py`, `collab.py`, `analytics.py`).

## Why

Since 403 wasn't in the retry branch, `raise_for_status()` fired immediately and the whole check/analytics call surfaced as a hard failure (a whole security check reported as `"error"`, or a whole cockpit/analytics endpoint failing) instead of transparently retrying like the 429 case already does.

## Fix

Added a predicate (duplicated in both files, matching how the retry loop itself is already duplicated with an explicit "same contract" comment) that treats a 403 as rate-limit-flavored **only** when it carries a `Retry-After` header or `X-RateLimit-Remaining: 0`. A genuine permission-denied 403 (token lacks the required scope) has neither header, so it's unaffected and still surfaces immediately — it wouldn't succeed on retry anyway, so retrying it would just waste two attempts and delay a response the caller can't do anything about faster.

## Test plan

- [x] `apps/api/tests/test_github_client.py` — new tests: retries on a 403 with `Retry-After`; retries on a 403 with `X-RateLimit-Remaining: 0`; does **not** retry a plain permission-denied 403 (asserts `time.sleep` not called, exactly 1 request attempt); same coverage for `request_paginated`.
- [x] `packages/checks/tests/test_github_checks.py` — new tests: `_get` retries on a rate-limit-flavored 403; does not retry a plain permission-denied 403.
- [x] `pytest -q` — 488 passed.

### DIFF
--- a/apps/api/src/services/github_client.py
+++ b/apps/api/src/services/github_client.py
@@ -5,6 +5,18 @@ from fastapi import HTTPException
 from src.core.config import settings
 
 
+def _is_secondary_rate_limit(resp: httpx.Response) -> bool:
+    """GitHub's secondary/abuse rate limit commonly returns 403 (not 429), often with a
+    Retry-After header -- especially under this codebase's concurrent ThreadPoolExecutor
+    fan-out (repos.py, collab.py, analytics.py). Without this, a 403 that would succeed on
+    retry raises immediately instead of backing off like the 429 case already does. A
+    genuine permission-denied 403 (e.g. token lacks scope) has neither header, so it's
+    unaffected and still surfaces immediately."""
+    if resp.status_code != 403:
+        return False
+    return "Retry-After" in resp.headers or resp.headers.get("X-RateLimit-Remaining") == "0"
+
+
 def github_error(exc: Exception) -> HTTPException:
     """Map an httpx exception raised by GitHubClient into the HTTPException every
     GitHub-proxying router returns to its caller."""
@@ -35,7 +47,7 @@ class GitHubClient:
                         time.sleep(2 ** attempt)
                         continue
                     raise
-                if resp.status_code == 429 and attempt < 2:
+                if (resp.status_code == 429 or _is_secondary_rate_limit(resp)) and attempt < 2:
                     time.sleep(2 ** attempt)
                     continue
                 resp.raise_for_status()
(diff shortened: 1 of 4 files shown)


nazarli-shabnam:
### Code review

No issues found. Verified the retry predicate is header-case-safe against real `httpx.Headers` (case-insensitive) in both files, the retry-condition change preserves exact prior 429 behavior with no off-by-one, and a plain permission-denied 403 (no rate-limit headers) still surfaces immediately rather than being retried away.

Confirmed downstream callers that pattern-match on 403 status (`BranchProtectionEnabled`, `DependabotAlertsCheck`, `CodeScanningCheck`) are unaffected — retrying only delays the eventual raise/response, it never changes the status code or exception type they see. This distinction between rate-limit-flavored and genuine permission-denied 403s is consistent with established precedent in this codebase (PR #127/#209's 403-vs-404, unknown-vs-fail handling), not a new pattern being introduced in isolation.

PR #240, opened 2026-07-23
fix(auth): concurrent /auth/register or /auth/setup calls can raise unhandled IntegrityError

### LINKED ISSUE #218: fix(auth): concurrent /auth/register or /auth/setup calls can raise unhandled IntegrityError
## Problem

`setup()` and `register()` in `apps/api/src/routers/auth.py:132-186` both do check-then-insert without catching `IntegrityError`:

- `setup()` (132-152): checks `db.query(User).count() > 0` then commits a new admin `User` row — no unique constraint stops two concurrent calls from both passing the count check and both becoming workspace admins.
- `register()` (155-186): checks `db.query(User).filter(User.email == body.email).first()` then commits — if two requests race with the same email, the second `db.commit()` raises `sqlalchemy.exc.IntegrityError` (from the presumed unique constraint on `email`), which isn't caught anywhere in the route, producing a raw 500.

This is inconsistent with the repo's own established pattern elsewhere (`org_repo.get_or_create`, `org_membership_repo.get_or_create`, `installation_repo.upsert`), which explicitly catch `IntegrityError` around the same class of race window.

## Failure scenario

Fire two concurrent `POST /auth/register` requests with the same email (e.g. a double-click or a retried request from a flaky client) — the second returns an unhandled 500 instead of a clean 409. Fire two concurrent `POST /auth/setup` requests with different emails before any user exists — both can succeed, creating two separate workspace admins.

## Suggested fix

Wrap the `db.commit()` in both routes in a `try/except IntegrityError`, `db.rollback()`, and return the appropriate 409, mirroring the existing get-or-create pattern used elsewhere in the codebase.

## Severity

Medium — reproducible with two simultaneous requests, easy single-PR fix mirroring an existing in-repo pattern.

nazarli-shabnam:
## What changed — this touches auth-sensitive code (`apps/api/src/routers/auth.py`), flagging per AGENTS.md

Fixes #218. `setup()` and `register()` in `apps/api/src/routers/auth.py` (lines ~132-186) both did check-then-insert without catching `IntegrityError`:

- **`register()`**: checks `db.query(User).filter(User.email == body.email)` then commits — if two requests race with the same email, the second `db.commit()` raised `sqlalchemy.exc.IntegrityError` (the unique constraint on `User.email`), unhandled anywhere, producing a raw 500.
- **`setup()`**: checks `db.query(User).count() > 0` then commits a new admin `User` row — no unique constraint stops two concurrent calls with *different* emails from both passing the count check and both becoming workspace admins.

This is inconsistent with the repo's own established pattern elsewhere (`org_repo.get_or_create`, `org_membership_repo.get_or_create`, `installation_repo.upsert`), which explicitly catch `IntegrityError` around the same class of race window.

## Fix

- **`register()`**: wraps `db.commit()` in `try/except IntegrityError: db.rollback()`, returning a clean 409 — mirrors the existing get-or-create pattern exactly, since `users.email` is genuinely unique and the second commit will collide for real.
- **`setup()`**: this one needed more than just try/except. The two-*different*-emails race has **no column collision** for a unique constraint to catch — both rows would insert successfully. Added a `pg_advisory_xact_lock` around the whole check-then-insert critical section instead: it serializes concurrent `/auth/setup` calls so the second call's `count()` check correctly sees the first call's already-committed row before attempting its own insert. Released automatically at commit/rollback — no schema change, no new dependency (Postgres builtin). Kept the `try/except IntegrityError` too, as defense-in-depth.

## Test plan

- [x] `apps/api/tests/test_auth.py::test_register_concurrent_same_email_returns_409_not_500` — same simulated-race pattern already used in `test_org_repo.py`'s concurrent-insert-race test (patches `Query.first` to fake a miss on the existence check, forcing the real insert into a genuine unique-constraint collision).
- [x] `apps/api/tests/test_auth.py::test_setup_advisory_lock_serializes_concurrent_holders` — directly verifies the lock primitive using two independent DB connections: a second connection can't acquire the same advisory-lock key while the first transaction holds it, and can once it's released.
- [x] `pytest -q` — 484 passed.

### DIFF
--- a/apps/api/src/routers/auth.py
+++ b/apps/api/src/routers/auth.py
@@ -142,8 +142,10 @@ def setup(body: SetupRequest, db: Session = Depends(get_db)):
     # count()==0 check before either commits -- unlike /auth/register's email race, there's
     # no unique-constraint collision to catch this (the two rows don't share a column
     # value), so it would otherwise silently create two workspace admins. The advisory
-    # lock serializes the whole check-then-insert critical section; it's released
-    # automatically when this transaction commits or rolls back.
+    # lock serializes the whole check-then-insert critical section; it's transaction-scoped,
+    # so it's released whenever this transaction ends -- an explicit commit/rollback below,
+    # or (on the early-409 path, which does neither) get_db()'s finally: db.close(), which
+    # implicitly rolls back and returns the connection to the pool at the end of the request.
     db.execute(text("SELECT pg_advisory_xact_lock(:key)"), {"key": _SETUP_LOCK_KEY})
     if db.query(User).count() > 0:
         raise HTTPException(status_code=409, detail="Setup already complete")
--- a/apps/api/src/routers/auth.py
+++ b/apps/api/src/routers/auth.py
@@ -21,6 +21,8 @@ import logging
 import bcrypt
 from fastapi import APIRouter, Depends, HTTPException, Response, status
 from pydantic import BaseModel, ConfigDict, EmailStr
+from sqlalchemy import text
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session
 
 from src.core.app_config import get_config
@@ -33,6 +35,10 @@ logger = logging.getLogger(__name__)
 router = APIRouter()
 
 _MIN_PASSWORD_LEN = 12
+
+# Arbitrary constant identifying the /auth/setup critical section for pg_advisory_xact_lock
+# (see setup()). Any int works; this one has no other significance.
+_SETUP_LOCK_KEY = 727100
 
 # Fixed hash checked when no real user/password_hash exists, so a login attempt for a
 # nonexistent email still pays the same bcrypt cost as a real one -- otherwise response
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

Found 1 issue (comment accuracy, not a functional bug), fixed in a follow-up commit:

1. The advisory-lock comment in `setup()` claimed the lock is "released automatically when this transaction commits or rolls back," but the early 409 path (`count() > 0`) does neither explicitly — it relies on `get_db()`'s `finally: db.close()`. Verified that path does correctly release the lock (`Session.close()` implicitly rolls back and returns the connection to the pool), so this was a comment-accuracy issue, not a bug. Fixed in [`b9e719a`](https://github.com/nazarli-shabnam/clevis/blob/b9e719aff8441441b317cc36a706933c221bb8c6/apps/api/src/routers/auth.py#L141-L149).

Otherwise clean — verified the lock is acquired before the check-then-insert window (not after), the `IntegrityError` catch in `register()` can't be triggered by anything but a genuine email collision, and both new tests' connection/lock handling is correct even on assertion failure (Python's `with`-block cleanup covers the untested path).

Worth noting for context: a prior PR (#207) explicitly identified this exact `/auth/setup` race and deliberately deferred fixing it, since the fix they considered (a partial unique index on `is_workspace_admin`) needed a migration, per AGENTS.md's migration guardrail. This PR's advisory-lock approach solves the same problem without a schema change — a better resolution than what was deferred, not something overlooking that prior context.

PR #241, opened 2026-07-23
feat(auth): add email verification, close invite-hijack gap

### LINKED ISSUE #217: fix(auth): unverified email registration lets an attacker hijack a pending org invite
## Problem

`POST /auth/register` (`apps/api/src/routers/auth.py:155-186`) creates an account with **no email-ownership verification** — the code's own comment (lines 182-184) explicitly acknowledges this: "register has no email-verification step, so a self-asserted email is not proof of inbox control."

Separately, `POST /invitations/{token}/accept` (`apps/api/src/routers/invitations.py:114-137`) grants org membership purely on a string match:

```python
if invitation.email.lower() != user.email.lower():
    raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, ...)
membership = org_membership_repo.get_or_create(db, org_id=invitation.org_id, user_id=user.id, role="member")
```

This is the **exact same attack class** that `apps/api/src/routers/github_auth.py`'s own module docstring (lines 9-18) explicitly reasons about and defends against for the GitHub-OAuth-linking path via `EmailAlreadyRegistered` — but the plain email/password registration path has no equivalent defense. It's an asymmetry between the app's two auth paths, not just an isolated missing check.

## Failure scenario

1. An org admin plans to (or already did) invite `victim@company.com` to their org.
2. An attacker calls `POST /auth/register` with `email="victim@company.com"` and a password the attacker controls — no verification email is sent, so this succeeds immediately.
3. When the invite is created/exists, the attacker (already authenticated as the account owning that email string) calls `POST /invitations/{token}/accept` (if they can obtain/guess the token, e.g. via a leaked invite link) and is granted `member` role in the org under the victim's email.
4. The real victim can no longer register that email (`register` now 409s "account already exists") and is locked out of their own invite, while the attacker silently holds org access tied to the victim's identity.

## Suggested fix

Require email verification before an account can accept an invitation (or before registration completes at all) — e.g. send a verification link on `register`, and gate `accept_invitation` on `email_verified = true`. At minimum, mirror the OAuth path's defense: refuse to let an unverified email/password account accept an invite for an email it hasn't proven ownership of.

## Severity

High — a real account-takeover-adjacent vector on the org invitation flow, symmetric with a risk the codebase already explicitly defends against on the OAuth path.

nazarli-shabnam:
## ⚠️ This PR includes a schema change AND touches auth-sensitive code — flagging both per AGENTS.md

Fixes #217. `POST /auth/register` created accounts with no email-ownership verification (the code's own comment already acknowledged this), and `POST /invitations/{token}/accept` granted org membership purely on a case-insensitive email string match. Combined, this let an attacker who knows a victim's email pre-register with it, then accept an invite meant for that email under the victim's identity — the exact attack class `github_auth.py`'s `EmailAlreadyRegistered` already defends against on the GitHub OAuth path (GitHub vouches for the email there), but the password path had no equivalent since there was nothing on the `User` model to represent "this email has been proven."

**User decision on scope**: after presenting three options (mirror the OAuth path's refuse-only defense / restrict invites to GitHub-linked accounts only / build a real email verification pipeline), the maintainer chose the full email verification pipeline, using plain SMTP (no third-party vendor) with SMTP config optional (graceful degrade if unset).

## Migration — why it was necessary

`0017_add_email_verification.py` adds `users.email_verified` (bool, NOT NULL), `email_verify_token`, `email_verify_token_expires_at` (both nullable). There is genuinely nothing to represent "has this email been proven" without a new column — additive-only, no data-loss risk. Existing rows are backfilled `email_verified=true` via `server_default` (already-deployed users are grandfathered in as trusted; this migration doesn't retroactively lock anyone out). New inserts set the value explicitly in application code. Ran `alembic revision --autogenerate` afterward to confirm zero drift between the model and the migration.

## What changed

- `users.email_verified`: `True` for GitHub-linked accounts (GitHub already verifies the email — `github_oauth.fetch_identity` only ever returns a GitHub-verified address) and the first-run `/auth/setup` admin (implicitly trusted, same reasoning as `is_workspace_admin=True` there). `False` for self-service `/auth/register` accounts until they click the emailed verification link.
- New `apps/api/src/services/email.py` sends the verification link via plain SMTP (stdlib `smtplib`, no new dependency). SMTP config is optional (`src.core.config.Settings`, same optional-block pattern already used for `github_app_*`) — if unset, `EmailNotConfigured` is raised and **registration proceeds anyway** with the account left unverified, logging a warning. Account creation must never depend on email sending working.
- New `POST /auth/verify-email` (token → marks verified, 24h expiry, rate-limited) and `POST /auth/resend-verification` (auth required, rate-limited per account).
- `invitations.py`'s `accept_invitation` now also requires `email_verified`, in addition to the existing email-match check, with a distinct 403 message the UI keys off of.
- UI: new `/verify-email` page (public route); the invite-accept page shows a "Resend verification email" button specifically when accept fails for this reason (distinct from a genuine email mismatch, which also 403s but with different wording). Registration's UI flow is otherwise unchanged — the account still logs the user in immediately; it just can't accept an org invite until verified.

## Test plan

- [x] `apps/api/tests/test_email_service.py` — new: raises `EmailNotConfigured` when SMTP unset; sends via SMTP with STARTTLS + login when credentials are set; sends without login when they aren't.
- [x] `apps/api/tests/test_auth.py` — new: setup creates a verified admin; register creates an unverified user with a pending token; register still succeeds when SMTP is unconfigured or email sending raises; verify-email happy path clears the token; verify-email rejects unknown/expired tokens; resend-verification regenerates the token / is a no-op for an already-verified account / requires auth.
- [x] `apps/api/tests/test_invitations.py` — new: accept is 403'd for an unverified account (distinct message), succeeds once verified. Updated the existing `_make_user` test helper to default `email_verified=True` so the pre-existing email-match tests keep testing exactly what they were testing before.
- [x] `apps/api/tests/test_github_auth.py` — new: a GitHub-created user is already verified.
- [x] `apps/ui/tests/components/verify-email-page.test.tsx`, `invite-accept-page.test.tsx` — new: success/error/missing-token states for the verify page; resend button appears only for the verification-specific 403, not other accept failures.
- [x] `pytest -q` — 513 passed.
- [x] `bun run test` — 286 passed (1 pre-existing unrelated flake in `repos-page.test.tsx`, confirmed passes in isolation).
- [x] `alembic upgrade head` applied cleanly; `alembic revision --autogenerate` afterward confirmed zero model/migration drift.

### DIFF
--- a/apps/api/alembic/versions/0017_add_email_verification.py
+++ b/apps/api/alembic/versions/0017_add_email_verification.py
@@ -1,0 +1,52 @@
+"""add users.email_verified/email_verify_token/email_verify_token_expires_at
+
+Issue #217: POST /auth/register creates an account with no email-ownership
+verification, and POST /invitations/{token}/accept grants org membership on a
+plain email string match -- so an attacker who knows a victim's email can
+register with it first, then accept an invite meant for that email under the
+victim's identity. The GitHub OAuth path already defends against the
+equivalent attack (EmailAlreadyRegistered refuses to auto-link on email
+match, since GitHub vouches for the email); the password path had no
+equivalent because there was nothing on the User model to represent "this
+email has been proven."
+
+- email_verified: bool, NOT NULL. Existing rows are backfilled true via
+  server_default=sa.true() -- every user who already has an account today is
+  grandfathered in as trusted (this migration doesn't retroactively lock
+  anyone out); going forward, application code explicitly sets it per user
+  (True for GitHub-linked/first-run-setup users, False for self-registered
+  users) rather than relying on the server default for new inserts.
+- email_verify_token / email_verify_token_expires_at: nullable, no backfill
+  needed -- only set when a verification email is pending.
+
+Additive-only, no data-loss risk. Same style as 0016_add_scan_results_scanned_by_user_id.py.
+
+Revision ID: 0017
+Revises: 0016
+Create Date: 2026-07-23
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "0017"
+down_revision = "0016"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "users",
+        sa.Column("email_verified", sa.Boolean(), nullable=False, server_default=sa.true()),
+    )
+    op.add_column("users", sa.Column("email_verify_token", sa.Text(), nullable=True))
+    op.add_column("users", sa.Column("email_verify_token_expires_at", sa.DateTime(timezone=True), nullable=True))
+    op.create_unique_constraint("uq_users_email_verify_token", "users", ["email_verify_token"])
+
+
+def downgrade() -> None:
+    op.drop_constraint("uq_users_email_verify_token", "users", type_="unique")
+    op.drop_column("users", "email_verify_token_expires_at")
+    op.drop_column("users", "email_verify_token")
+    op.drop_column("users", "email_verified")
--- a/apps/api/src/core/config.py
+++ b/apps/api/src/core/config.py
@@ -41,6 +41,16 @@ class Settings(BaseSettings):
     session_cookie_samesite: str = "lax"
     session_cookie_domain: str | None = None
 
+    # SMTP (email verification, issue #217). Optional -- same pattern as github_app_* above:
+    # if unset, src.services.email raises a clear "not configured" error only when actually
+    # used, and the register() route degrades gracefully (account still created, just stays
+    # unverified) rather than requiring every deployment to have working email.
+    smtp_host: str | None = None
+    smtp_port: int = 587
+    smtp_user: str | None = None
+    smtp_password: SecretStr | None = None
+    smtp_from: str | None = None
+
     model_config = SettingsConfigDict(env_file=_env_file, extra="ignore")
 
 
--- a/apps/api/src/core/db.py
+++ b/apps/api/src/core/db.py
@@ -116,6 +116,14 @@ class User(Base):
     github_login: Mapped[str | None] = mapped_column(Text, nullable=True)
     avatar_url: Mapped[str | None] = mapped_column(Text, nullable=True)
     created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now())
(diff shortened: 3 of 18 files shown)


nazarli-shabnam:
### Code review (high effort — migration + auth-sensitive code)

Found 2 issues, both fixed in a follow-up commit:

1. **Stale comment**: `PendingInvitationSummary`'s comment still said "registration has no email-verification step anywhere in this app" — stale, since this same PR adds exactly that. Reworded to match `register()`'s already-correct comment (the email isn't verified *yet* at that point, not never).
2. **Real bug**: `verify_url` construction (`f"{settings.cors_origins[0]}/..."`) sat outside the try/except in `_send_verification_email_best_effort`. An operator misconfiguring `CORS_ORIGINS=[]` (empty array) would raise an uncaught `IndexError` there, breaking **registration itself**, not just degrading the email send — contradicting the function's own "never raises" docstring.

Both fixed in [`ecea3ff`](https://github.com/nazarli-shabnam/clevis/blob/ecea3ffbabc48071eaf5b4c086f5d8ef72d7a441/apps/api/src/routers/auth.py#L51-L67), with a new regression test (`test_register_still_succeeds_when_cors_origins_is_misconfigured_empty`) confirming registration now survives that misconfiguration.

**Migration**: verified genuinely necessary (no existing column represents "email ownership proven"), additive-only, `alembic revision --autogenerate` confirmed zero drift after applying. One non-blocking note: the SQLAlchemy model's ORM-level `default=False` and the migration's `server_default=sa.true()` intentionally diverge (existing rows grandfathered true, new inserts explicit in code) — every current insert path (`setup`, `register`, `find_or_create_user`) does set it explicitly, so not a bug, but worth remembering for any future insert path.

**Security**: traced the full attack end-to-end — token generation (`secrets.token_urlsafe(32)`, single-use, tz-aware expiry check) is sound, and the `accept_invitation` fix re-queries the DB by `user.id` rather than trusting JWT claims, correctly closing the hijack. No SQL injection, IDOR, or race-condition issues found.

**Context**: git history shows this exact gap was named as the root cause in three separate prior commits (`56ec2cf`, `c0239be`, `0219f12`) but worked around each time rather than closed — this PR is the first to actually close it.

PR #242, opened 2026-07-23
fix(worker): stale-job reclaim can duplicate execution and mislabel successful jobs as failed

### LINKED ISSUE #215: fix(worker): stale-job reclaim can duplicate execution and mislabel successful jobs as failed
## Problem

`apps/worker/src/worker.py:_reclaim_stale_jobs()` (lines 179-207) resets any job stuck in `processing` for longer than `RECLAIM_TIMEOUT_MINUTES` (30 min, line 28) back to `queued`, assuming the original worker crashed. There is no lease-renewal/heartbeat mechanism for jobs that are simply slow (e.g. a sluggish GitHub API response) rather than crashed.

`_mark_done`/`_mark_failed` (lines 63-82) do guard their `UPDATE` with `WHERE status='processing'`, which prevents a *lost update* — but it does not prevent the underlying race.

## Failure scenario

1. Worker A claims job N and starts a slow GitHub API call (e.g. cache clear) that legitimately takes >30 minutes.
2. `_reclaim_stale_jobs` (running in the poll loop of any worker, including A or B) resets job N to `queued` and increments `retry_count`.
3. Worker B's `SELECT ... FOR UPDATE SKIP LOCKED` picks up job N and starts the same GitHub side effect again — duplicate `DELETE .../actions/caches` call.
4. If B finishes first and marks the job `failed` (e.g. exceeding `MAX_RETRIES` via the reclaim's own retry_count bump) before A's original call completes, the job is permanently recorded `failed` in the audit trail even though A's call actually succeeded — a silent correctness gap in the job/audit history.

## Suggested fix

Add a heartbeat/lease-renewal column that a long-running handler updates periodically (e.g. touch `updated_at` mid-flight), and have `_reclaim_stale_jobs` only reclaim jobs whose heartbeat is stale, not just their original `updated_at`.

## Severity

High — breaks the job queue's core exactly-once-ish guarantee and can produce duplicate external side effects plus an inaccurate audit trail.

nazarli-shabnam:
## ⚠️ This PR includes a schema change — flagging per AGENTS.md

Fixes #215. `apps/worker/src/worker.py`'s `_reclaim_stale_jobs()` resets any job stuck in `processing` for longer than `RECLAIM_TIMEOUT_MINUTES` (30 min) back to `queued`, assuming the original worker crashed. There is no heartbeat/lease-renewal for jobs that are simply slow (e.g. a sluggish GitHub API response) rather than crashed — `updated_at` is only set once, at claim time, so on its own it can't tell a legitimately slow job apart from a crashed one once that window elapses.

## Migration — why it was necessary

`0017_add_jobs_heartbeat_at.py` adds `jobs.heartbeat_at` (nullable timestamptz, no backfill needed, additive-only). There's no existing column that can represent "still alive as of N seconds ago" for a running job. **Note**: main only had migrations through `0016` when this branch was created — if another in-flight PR's migration lands as `0017` first, this one will need renumbering to `0018` and rebasing at merge time.

## Failure scenario this closes

1. Worker A claims job N and starts a legitimately slow call.
2. The reclaim sweep (running in any worker's poll loop) resets job N to `queued` since `updated_at` looks stale.
3. Worker B's `SELECT ... FOR UPDATE SKIP LOCKED` picks up job N and starts the same external side effect again — a duplicate GitHub API call.
4. If B finishes first and marks the job `failed` (e.g. exceeding `MAX_RETRIES` via the reclaim's own `retry_count` bump) before A's original call completes, the job is permanently recorded `failed` in the audit trail even though A's call actually succeeded.

**Honesty about current impact**: the one existing job handler (`_handle_clear_actions_cache`, a single `httpx` call with a 20s timeout) realistically finishes in ~20-30s — nowhere near the 30-minute reclaim window, so this isn't an active bug against it today. It's a real latent gap for any future longer-running handler, and the fix matches the issue's own suggested approach, so I implemented it now per the explicit request rather than waiting for a handler that actually triggers it.

## Fix

A new `_JobHeartbeat` context manager runs on a background thread (its own DB connection, separate from the one `process_job` uses on the main thread — psycopg connections aren't safe to share across threads) and touches `jobs.heartbeat_at` every 10s for as long as a handler is running, wrapping `process_job`'s call in `run()`. The reclaim sweep's SQL now also requires `heartbeat_at` to be stale (or `NULL`, for a job claimed before this column existed) before reclaiming, so a genuinely-still-running job's fresh heartbeat protects it past 30 minutes.

## Test plan

- [x] `apps/worker/tests/test_reclaim.py` — new tests: a stale-`updated_at` job with a fresh `heartbeat_at` is left alone; one with a stale `heartbeat_at` (or `NULL`, for pre-existing jobs) is still reclaimed as before.
- [x] `apps/worker/tests/test_job_heartbeat.py` — new: `_touch_job_heartbeat` sets the column on a `processing` job and no-ops for a job that's no longer processing; `_JobHeartbeat` ticks repeatedly while its context is open (short interval override so the test doesn't take the real 10s) and stops ticking after it exits.
- [x] `pytest -q` — 503 passed.
- [x] `alembic upgrade head` applied cleanly; `alembic revision --autogenerate` afterward confirmed zero model/migration drift.

### DIFF
--- a/apps/api/alembic/versions/0017_add_jobs_heartbeat_at.py
+++ b/apps/api/alembic/versions/0017_add_jobs_heartbeat_at.py
@@ -9,7 +9,7 @@ duplicate its external side effect, and depending on timing the job can end
 up mislabeled 'failed' even though the original attempt actually succeeded.
 
 heartbeat_at is touched by the worker every ~10s while a job handler is
-actually running (apps/worker/src/worker.py's _heartbeat_while_running), so
+actually running (apps/worker/src/worker.py's _JobHeartbeat / _touch_job_heartbeat), so
 _reclaim_stale_jobs can check it instead of/alongside updated_at and leave a
 genuinely-still-running job alone past 30 minutes.
 
--- a/apps/api/alembic/versions/0017_add_jobs_heartbeat_at.py
+++ b/apps/api/alembic/versions/0017_add_jobs_heartbeat_at.py
@@ -1,0 +1,38 @@
+"""add jobs.heartbeat_at
+
+Issue #215: _reclaim_stale_jobs resets any job stuck in 'processing' for
+longer than RECLAIM_TIMEOUT_MINUTES (30 min) back to 'queued', assuming its
+worker crashed -- but updated_at is only set once, at claim time, so a
+legitimately slow (not crashed) job looks identical to a crashed one once
+that window elapses. A second worker can then pick up the same job and
+duplicate its external side effect, and depending on timing the job can end
+up mislabeled 'failed' even though the original attempt actually succeeded.
+
+heartbeat_at is touched by the worker every ~10s while a job handler is
+actually running (apps/worker/src/worker.py's _heartbeat_while_running), so
+_reclaim_stale_jobs can check it instead of/alongside updated_at and leave a
+genuinely-still-running job alone past 30 minutes.
+
+Nullable, no backfill needed -- additive-only. Same style as
+0014_add_jobs_retry_count.py.
+
+Revision ID: 0017
+Revises: 0016
+Create Date: 2026-07-23
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "0017"
+down_revision = "0016"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column("jobs", sa.Column("heartbeat_at", sa.DateTime(timezone=True), nullable=True))
+
+
+def downgrade() -> None:
+    op.drop_column("jobs", "heartbeat_at")
--- a/apps/api/src/core/db.py
+++ b/apps/api/src/core/db.py
@@ -85,6 +85,13 @@ class Job(Base):
     retry_count: Mapped[int] = mapped_column(Integer, nullable=False, default=0)
     created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now())
     updated_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), onupdate=func.now())
+    # Touched by the worker every ~10s while a job handler is actually running (see
+    # apps/worker/src/worker.py's _heartbeat_while_running). Lets _reclaim_stale_jobs tell
+    # a genuinely slow-but-alive job apart from one whose worker crashed -- updated_at alone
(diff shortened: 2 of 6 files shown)


nazarli-shabnam:
### Code review

Found 2 issues, both fixed in a92824d:

1. The new `jobs.heartbeat_at` mechanism didn't sync with the container-level file heartbeat the worker's Docker healthcheck relies on. `run()`'s poll loop only touches `HEARTBEAT_FILE` once per iteration, before a job is even claimed (https://github.com/nazarli-shabnam/clevis/blob/a92824d3091c482cbd4552fa25a5baf42e4eb5f3/apps/worker/src/worker.py#L292-L320), while the healthcheck treats the file stale after 60s (https://github.com/nazarli-shabnam/clevis/blob/a92824d3091c482cbd4552fa25a5baf42e4eb5f3/docker-compose.yml#L74-L82). A job handler running long enough to need this PR's DB heartbeat protection -- exactly the "future longer-running handler" scenario the PR describes -- would get the worker marked unhealthy mid-job even though it's legitimately alive, since the file heartbeat goes stale while the DB heartbeat keeps ticking. Fixed by also touching `HEARTBEAT_FILE` from `_touch_job_heartbeat`, so both heartbeats agree: https://github.com/nazarli-shabnam/clevis/blob/a92824d3091c482cbd4552fa25a5baf42e4eb5f3/apps/worker/src/worker.py#L242-L262

2. Two comments referenced a `_heartbeat_while_running` symbol that doesn't exist anywhere in `worker.py` -- leftover from an earlier draft name; the actual symbols are `_JobHeartbeat` / `_touch_job_heartbeat`. Fixed in the migration docstring (https://github.com/nazarli-shabnam/clevis/blob/a92824d3091c482cbd4552fa25a5baf42e4eb5f3/apps/api/alembic/versions/0017_add_jobs_heartbeat_at.py#L10-L15) and the `Job.heartbeat_at` column comment (https://github.com/nazarli-shabnam/clevis/blob/a92824d3091c482cbd4552fa25a5baf42e4eb5f3/apps/api/src/core/db.py#L88-L94).

Added a regression test for #1 (`test_touch_job_heartbeat_also_refreshes_the_container_heartbeat_file`). All 13 worker heartbeat/reclaim tests pass.

PR #254, opened 2026-07-24
fix(ui): Activity page required a legacy PAT despite GitHub App auth

### LINKED ISSUE #251: Activity page requires a legacy PAT, ignoring the GitHub-App-installation token fallback
## Summary

The Activity page (`apps/ui/app/activity/page.tsx:60-130`) gates every data query on a `configured` flag derived solely from whether a legacy saved PAT resolves:

```js
const resolveQuery = useQuery({
  queryKey: ["tokens.resolve", org],
  queryFn: () => api.tokens.resolve(org),
  enabled: org.trim().length > 0,
  retry: false,
})

const configured = !!resolveQuery.data?.token
const token = resolveQuery.data?.token ?? ""

const eventsQuery = useQuery({
  queryKey: ["github.events", org],
  queryFn: () => {
    if (!token) throw new Error("No GitHub token available for this organization")
    return api.github.events(org, token)
  },
  enabled: configured,
  ...
})
```

`api.tokens.resolve` calls `POST /tokens/resolve` (`apps/api/src/routers/tokens.py:83-98`), which only looks up the legacy `saved_tokens` table:

```python
@router.post("/resolve", response_model=VerifyTokenResponse)
def resolve_token(body, db, user=Depends(require_workspace_admin)):
    row = db.query(SavedToken).filter_by(org=body.org).first()
    if not row:
        raise HTTPException(status_code=404, detail="No saved token for this org")
    ...
```

There is **no fallback to a connected GitHub App installation token** in this lookup. Every query on the Activity page — `eventsQuery`, `cockpitQuery`, `failedRunsQuery`, `releaseTimelineQuery`, `reposQuery`, `prBoardQuery` — is gated on `configured`, and `eventsQuery`'s own `queryFn` throws before even attempting a request if there's no PAT.

## Failure scenario

An org connected purely via "Install GitHub App" (no legacy PAT ever saved — this is a fully supported and actively encouraged connection method, per the Settings page install flow) opens the Activity page and sees: *"No organization configured yet. Configure →"* — permanently, even though the org is fully connected and working. The feed, PR board, CI failure log, release timeline, and commit heatmap never load for that org.

This contradicts the pattern used elsewhere: on the Overview page (`apps/ui/app/page.tsx`), `cockpitQuery`/`myViewQuery` are enabled based on `org.trim().length > 2 && !resolveQuery.isLoading` — independent of whether a PAT actually resolved — because the underlying API endpoints for Overview and Security already fall back to a GitHub App installation token server-side (see `token_resolution.py`'s `resolve_org_token`). Activity is the outlier that hard-requires a PAT client-side.

## Why this matters

This silently breaks a core page for any org onboarded via the GitHub-App-only flow (the flow the product actively steers people toward, since it avoids browser-pasted PATs entirely). A user could reasonably conclude the whole product is broken for their org, when only this one page has the wrong gating condition.

Change the Activity page's query-enablement to match the Overview page's pattern — enable queries once an org is selected (not conditioned on `resolveQuery.data?.token`), and pass `token || undefined` to the API calls that already support the installation-token fallback server-side, the same way `api.github.events` is documented to support (`token: token || undefined` per `lib/api/client.ts`).

nazarli-shabnam:
## Summary

Closes #251.

The Activity page (`apps/ui/app/activity/page.tsx`) derived a `configured` flag from whether a legacy PAT resolved via `POST /tokens/resolve`, and gated **every** query on the page (event feed, commit heatmap, CI failure log, release timeline, PR board) behind that flag. If no PAT had ever been saved for the org — a fully supported setup when the org is connected purely via GitHub App installation — `configured` stayed `false` forever, and the page showed a permanent "No organization configured yet" prompt even though the org was correctly connected and Overview/Security worked fine for it.

Overview (`apps/ui/app/page.tsx`) already avoids this: its data queries are enabled once an org is known and token resolution has settled, regardless of whether a PAT was actually found, trusting the API's own installation-token fallback (`token_resolution.resolve_org_token`/`resolve_personal_token`). Activity never got that same treatment.

## What changed

- Replaced the `configured` (PAT-resolved) gate with `hasOrg` (an org is set) for every query's `enabled` flag, matching Overview's pattern.
- The "no organization" empty state now reflects whether an org is actually configured, not whether a PAT resolved.
- Added a `SectionError` (with retry) branch for the event feed specifically, since a genuine "no token available at all" failure now surfaces as a real query error instead of being silently swallowed by the old gate.
- Updated/added tests in `apps/ui/tests/components/activity-page.test.tsx`: the old test asserting the buggy behavior (no PAT → events never queried) is replaced with a test asserting the fixed behavior (no PAT → events still queried, relying on the API's installation-token fallback), plus a new test for the genuine-no-token error path.

## Why

An org connected only via GitHub App installation (no PAT ever saved) is meant to be a fully working setup per this project's architecture — Activity being blank for that case undermines the whole point of the GitHub App migration.

No RBAC/auth/token-encryption/migration changes — this is a client-side query-gating fix only; the backend token-resolution fallback it now relies on already existed and is unchanged.

## Test plan
- [x] `bun run test -- activity-page` — all 10 tests pass
- [x] `bun run typecheck` — passes
- [ ] CI

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Activity page GitHub/analytics loading for organizations without a saved token, ensuring queries run after token resolution completes.
  - Activity feed now handles missing/empty tokens gracefully and shows the “no events yet” state instead of incorrect prompts.
  - Added a clearer error experience on event load failures, including a **Retry** action to recover without refresh.
  - Refined when refresh countdown and right-side activity panels (e.g., heatmap/CI/release timeline) appear, based on org availability.
  - Tightened pull request board querying to avoid incomplete data states.
- **Tests**
  - Updated Activity page coverage for empty-token and error/retry scenarios.

### DIFF
--- a/apps/ui/app/activity/page.tsx
+++ b/apps/ui/app/activity/page.tsx
@@ -7,6 +7,7 @@ import { PageHeader } from "@/components/page-header"
 import { ActivityList } from "@/components/activity-list"
 import { EventFeed } from "@/components/event-feed"
 import { HeatmapCalendar } from "@/components/charts/heatmap-calendar"
+import { SectionError } from "@/components/section-error"
 import { CHART_COLORS } from "@/lib/charts/theme"
 import { relativeTime } from "@/lib/format"
 import { api } from "@/lib/api/client"
@@ -64,16 +65,20 @@ export default function ActivityPage() {
     retry: false,
   })
 
-  const configured = !!resolveQuery.data?.token
   const token = resolveQuery.data?.token ?? ""
+  // Queries fire once an org is set and token resolution has settled, regardless of
+  // whether a saved PAT was found -- an org connected purely via GitHub App
+  // installation (no PAT ever saved, a fully supported flow) has no `token` here, but
+  // the API resolves an installation token server-side the same way Overview's queries
+  // already rely on (see app/page.tsx). Gating on a resolved PAT specifically made this
+  // page permanently blank for App-only orgs (#251).
+  const hasOrg = org.trim().length > 0
+  const queriesEnabled = hasOrg && !resolveQuery.isLoading
 
   const eventsQuery = useQuery({
     queryKey: ["github.events", org],
-    queryFn: () => {
-      if (!token) throw new Error("No GitHub token available for this organization")
-      return api.github.events(org, token)
-    },
-    enabled: configured,
+    queryFn: () => api.github.events(org, token),
+    enabled: queriesEnabled,
     retry: false,
     refetchInterval: EVENTS_REFRESH_SECONDS * 1000,
   })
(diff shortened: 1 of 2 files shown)


nazarli-shabnam (commented):
Confirmed this fully addresses #251 — Activity's queries now key off org selection instead of a resolved PAT, matching Overview's pattern, and the new tests cover both the empty-token and genuine-error cases. CI green.

PR #255, opened 2026-07-24
fix(ui): audit tab's error state hides an unrelated, loaded section

### LINKED ISSUE #252: Permission-audit tab's error state hides an unrelated, successfully-loaded section
## Summary

The org members page's Audit tab (`apps/ui/app/settings/org/[login]/members/page.tsx`) renders two independently-fetched sections — a permission audit and an "Inactive members (30d+)" list (lines 340-381) — but the shared top-level loading/error gate only tracks the permission-audit query:

```js
const activeQuery =
  tab === "members" ? membersQuery
  : tab === "outside" ? outsideQuery
  : tab === "pending" ? pendingQuery
  : permissionAuditQuery   // used for tab === "audit"
...
) : activeQuery.isError ? (
  <p className="text-xs text-destructive">{activeQuery.error.message}</p>
) : ...
```

(lines 84-88, 164-171)

## Failure scenario

An org admin opens the Audit tab. `permissionAuditQuery` fails — e.g. GitHub rate-limits or 403s that specific endpoint — while the separately-fetched `inactiveMembersQuery` succeeds and has data ready to show. Because the top-level gate only checks `activeQuery.isError` (== `permissionAuditQuery.isError`), the whole panel collapses to a single error line, and the "Inactive members (30d+)" section — which loaded successfully and has real data — is never rendered. The admin loses visibility into stale/inactive members that Clevis actually has data for, purely because of an unrelated failure in a different query.

## Why this matters

This is a UX correctness bug: partial data that's genuinely available gets hidden behind an error for a query the admin doesn't even care about in that moment. It also makes debugging harder — an admin sees a rate-limit-flavored error message and has no way to tell that it only affects half the tab's content.

Gate each section's rendering on its own query's loading/error state independently, rather than sharing a single `activeQuery` derived value for the Audit tab specifically (the other three tabs, which only render one section each, are unaffected by this and don't need the change).

nazarli-shabnam:
## Summary

Closes #252.

`apps/ui/app/settings/org/[login]/members/page.tsx`'s Audit tab renders two independently-fetched sections -- the permission audit table and "Inactive members (30d+)" -- but the top-level loading/error gate (`activeQuery.isLoading` / `activeQuery.isError`) only ever tracked `permissionAuditQuery`. If the permission-audit call failed (e.g. a GitHub rate limit hitting that specific endpoint) while the inactive-members call succeeded, the whole tab collapsed to a single generic error line, and inactive-member data that loaded fine was never shown to the admin.

## What changed

- The top-level `isLoading`/`isError` gate now excludes the `audit` tab (`tab !== "audit" && ...`) -- it was never a good fit for a tab backed by two independent queries.
- The permission-audit table gets its own `isLoading`/`isError` branch, mirroring the handling `inactiveMembersQuery` already had further down in the same tab.
- Added a test (`apps/ui/tests/components/github-roster.test.tsx`) covering the actual failure scenario from the issue: permission audit fails, inactive members succeeds -- both the error message and the successfully-loaded inactive member now render simultaneously.

## Why

An admin using the Audit tab for a real security review shouldn't lose visibility into inactive members just because an unrelated GitHub API call (permission audit) hit a rate limit or transient error.

No RBAC/auth/token-encryption/migration changes -- purely a client-side rendering fix; no backend changes.

## Test plan
- [x] `bun run test -- github-roster` -- all 18 tests pass
- [x] `bun run typecheck` -- passes
- [ ] CI

### DIFF
--- a/apps/ui/app/settings/org/[login]/members/page.tsx
+++ b/apps/ui/app/settings/org/[login]/members/page.tsx
@@ -161,11 +161,15 @@ function GithubRoster({ orgLogin }: { orgLogin: string }) {
         )}
       </div>
 
-      {activeQuery.isLoading ? (
+      {/* The audit tab renders two independently-fetched sections (permission audit,
+          inactive members) and gives each its own loading/error state below -- gating the
+          whole tab on permissionAuditQuery alone would hide inactive-members data that
+          loaded fine just because the audit call happened to fail (#252). */}
+      {tab !== "audit" && activeQuery.isLoading ? (
         <div className="px-4 py-6 flex items-center gap-2 text-sm text-muted-foreground">
           <CircleNotch className="size-3.5 animate-spin" /> Loading…
         </div>
-      ) : activeQuery.isError ? (
+      ) : tab !== "audit" && activeQuery.isError ? (
         <div className="px-4 py-6">
           <p className="text-xs text-destructive">{activeQuery.error.message}</p>
         </div>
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
Confirmed — the audit tab now gates the permission-audit table and inactive-members list independently, so one section's error no longer hides the other's loaded data. Clean, minimal fix for #252, test covers the exact scenario. CI green. Thanks!

PR #205, opened 2026-07-17
fix(ui): drop legacy PAT inputs from Security/Cache/Overview

Tarekchehahde:
### Why
Issue #189: GitHub App installation tokens are live server-side (`token_resolution`), but Health & Security / Cache / Overview still behaved like a PAT-first product — paste fields, `/tokens/resolve`, and README step “add a PAT”. That undermined the App migration and left users pasting long-lived tokens into the browser.

### What changed (and why each piece)
- **Security + Cache UI** — Removed PAT input, save-token, and `/tokens/resolve` auto-fill. Pages call the API with no client token so the backend mints an installation token. Help copy points at Settings → Connected orgs when requests fail.
- **Overview** — Stopped gating stats on a resolved saved PAT. Uses App path with `default_org`. **Configure →** now goes to `/settings` (App install), not `/security`. Failed overview calls show the API error + Settings CTA instead of silent `"—"`.
- **API client** — Omits empty/whitespace tokens so a blank field cannot poison App-first resolution. Explicit tokens still work for tools/tests.
- **`token_resolution` errors** — Message now steers to Connected orgs (App-first), not “paste a PAT”.
- **Settings copy** — Legacy PAT section described as fallback-only.
- **README Get Started** — Dropped the required-PAT step.

### Sensitive surfaces (AI attribution policy)
Touches **token-handling UX** and `token_resolution` user-facing errors (not encryption keys / Fernet / RBAC deps). No migrations. No auth dependency changes.

### Not in this PR
Removing the Settings legacy PAT storage APIs entirely (still a valid server-side fallback if a client token is supplied).

Fixes #189

## Test plan
- [x] `cd apps/ui && bun run test -- tests/components/overview-page.test.tsx tests/components/settings-page.test.tsx tests/components/security-page.test.tsx tests/components/cache-page.test.tsx tests/lib/client.test.ts` (33 passed)
- [x] `cd apps/ui && bun run typecheck`
- [ ] CI UI + E2E green
- [ ] Manual: App connected → scan/load caches with no PAT
- [ ] Manual: no App + `default_org` set → Overview shows error + Settings link

PR #256, opened 2026-07-24
docs(readme): drop stale note that a legacy token is still required

### LINKED ISSUE #189: feat(ui): Security/Cache pages still require a legacy PAT despite GitHub App auth being live
## Problem

GitHub App installation-token auth is live (S1, closed via #140/#56), but `README.md:94` still documents adding a "personal access token (temporary, still required)" because "Health & Security and Cache pages haven't moved onto the GitHub App token path yet." This matches `docs/plan.md:99-100`'s explicit "⬜ Remaining" note: "drop the legacy PAT inputs from Security/Cache pages (with S6)."

## User impact

Every user still has to manually create and paste a GitHub PAT into Settings to use the two most security-relevant pages in the product, even though the whole point of S1's GitHub App migration was to eliminate PAT-in-the-browser entirely. This is the single biggest visible gap between "GitHub App auth is done" and "PAT-free" being actually true for the end user.

## Suggested fix

Wire `apps/ui/app/security/page.tsx` and the cache pages (`apps/ui/app/repos/[repo]/cache/page.tsx`) to resolve and use the installation token (already implemented server-side per #140) instead of falling back to a saved PAT via `/tokens/resolve`. Note: `docs/plan.md` ties this to S6 (aggregates-first API), but the token-resolution swap itself doesn't structurally require S6 — it can likely be done as soon as the installation-token service is trusted for read-only GitHub API calls, which it already is per #140's resolution.

## Severity

Medium — directly user-visible, and undermines the security pitch of the GitHub App migration until it's fixed.

nazarli-shabnam:
## Summary

Closes #189.

#189 claimed "Every user still has to manually create and paste a GitHub PAT into Settings to use the two most security-relevant pages" (Health & Security, Cache). Investigating this turned up that it's already false on `main`:

- `apps/ui/app/security/page.tsx` and `apps/ui/components/repo/cache-panel.tsx` both auto-resolve a saved PAT via `/tokens/resolve` but treat it as fully optional (`token || undefined` sent to the API either way), already labeled "optional if the GitHub App is connected for this org" in the UI.
- The API-side installation-token fallback (`token_resolution.resolve_org_token`/`resolve_personal_token`) that makes this work was wired into these endpoints back in `dc889ae` (`fix(api): wire GitHub App installation tokens into analytics and cache endpoints`), well before this issue was filed.

So the only thing actually stale was this README line, which still framed adding a PAT as a required numbered setup step with a "temporary" caveat and a link tracking its removal.

## What changed

- `README.md`: removed the numbered "(Temporary) Add a personal access token only if needed" step and its `#189` tracking link; folded the PAT note into the closing sentence as what it actually is — an optional fallback for orgs without a GitHub App installation, not a setup action.

## Why not adopt PR #205 instead

There's an existing open PR (#205) also targeting #189 that goes further: it removes the manual PAT input field from these pages entirely, making them GitHub-App-only. I did not adopt that approach here, because it would be a behavior regression, not just a docs fix — `resolve_org_token`/`resolve_personal_token` only fall back to a legacy PAT when the **UI itself** supplies one in the request body; there's no server-side auto-lookup of `saved_tokens`. Removing the input field would silently break Security/Cache for any org connected via a legacy-PAT-only setup with no GitHub App installed (a still-supported configuration per this project's docs). That's a larger, separate decision than what #189 as filed actually needed.

No RBAC/auth/token-encryption/migration changes — README only, no code changes.

## Test plan
- [x] N/A — documentation-only change
- [ ] CI

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,9 +91,8 @@ Once the instance is running, here's how to go from a fresh deploy to a connecte
 
 1. **Create the first account.** The first visit lands on `/setup`, which creates the initial admin. After that, sign in with a password or the "Sign in with GitHub" button.
 2. **Connect a GitHub org or account.** Go to **Settings → Connected orgs** and click **Install GitHub App** — this sends you to GitHub to authorize Clevis for that org/account, which is what grants read access to its repos. GitHub redirects you back into the app once you approve it, and the org/account shows up under "Connected orgs" a few seconds later. (The App itself is a one-time setup step for whoever deployed the instance — see [`docs/self-hosting.md`](docs/self-hosting.md) if that hasn't been done yet.) If you install into a brand-new org and the connection doesn't complete, sign out and back in with **"Sign in with GitHub"** once (this is what verifies your admin access on that org) and retry the install.
-3. **(Temporary) Add a personal access token only if needed.** Health & Security and Cache still accept a legacy PAT via **Settings → Personal access tokens (legacy)** as a fallback when no GitHub App installation is available for that org. Tracking removal of that requirement: [#189](https://github.com/nazarli-shabnam/clevis/issues/189).
 
-That's it — Overview, Activity, Repositories, Collaborators, and (when the App is connected) Health & Security / Cache work off the connected org from step 2.
+That's it — Overview, Activity, Repositories, Collaborators, Health & Security, and Cache all work off the connected org from step 2. A legacy personal access token (**Settings → Personal access tokens (legacy)**) is only ever needed as a manual fallback for an org that has no GitHub App installation — it is not a required step.
 
 ---
 


nazarli-shabnam:
Verified the premise directly: security/page.tsx and cache-panel.tsx already label the token field as optional ("optional if the GitHub App is connected for this org") and send it through as-is, letting the API's installation-token fallback resolve it — so #189 really was stale by the time this was filed. README now matches reality. One thing to sort out before merging: this overlaps with #205, which takes a more invasive approach to the same issue — worth deciding which one closes #189 so they don't step on each other. CI green otherwise. Thanks!

PR #257, opened 2026-07-24
fix(api): case-sensitive installation lookup can break token resolution

### LINKED ISSUE #246: Case-sensitive installation lookup can break GitHub App token resolution
## Summary

`installation_repo.get_for_org` / `get_for_user` (`apps/api/src/repositories/installation_repo.py:86-102`) do an exact `==` match on `account_login`:

```python
def get_for_org(db: Session, org_id: int, account_login: str) -> GitHubInstallation | None:
    return (
        db.query(GitHubInstallation)
        .filter(GitHubInstallation.org_id == org_id, GitHubInstallation.account_login == account_login)
        .first()
    )
```

`account_login` is stored verbatim from the GitHub install payload (`apps/api/src/routers/installations.py:165, 207`). But every ownership/RBAC check elsewhere in the codebase compares login casing **case-insensitively**:

- `assert_owner_matches_org` — `apps/api/src/core/rbac.py:52`: `if owner.lower() != ctx.org.github_login.lower(): ...`
- `_verify_installation` — `apps/api/src/routers/installations.py:72`: `if actual_login.lower() != account_login.lower() ...`
- `sync_org_installation` — `apps/api/src/routers/installations.py:137`: `if payload.account_login.lower() != org_login.lower(): ...`

## Failure scenario

Any endpoint that resolves a token by calling `resolve_org_token`/`resolve_personal_token` with an `owner`/`org_login` path parameter (e.g. `apps/api/src/routers/repos.py`, `github.py`, `automation.py`, `analytics.py`, `collab.py`, `security.py`, `actions_cache.py` — all pass `account_login=owner` straight from the request) can hit this: if the casing of that path parameter differs even slightly from what's stored in `github_installations.account_login` (org renamed casing on GitHub since install, or a client simply sends `"Acme"` instead of `"acme"`), the RBAC/ownership check in `rbac.py` passes (case-insensitive), but `get_for_org`/`get_for_user` return `None` because the exact-match query misses. `resolve_org_token` then either silently falls back to a client-supplied PAT (if one happens to be provided) or raises `NoGitHubTokenAvailable`, telling the user "No GitHub App installation found" even though one exists in the DB with different casing.

## Why this matters

This is directly adjacent to the recently-fixed "connecting a GitHub org can silently leave Overview empty" bug (see PR for issue tracked in this repo's history, branch `fix/github-app-install-org-not-connecting`). That fix didn't address casing — this is a related edge case in the same code path that could reintroduce a similar silent-failure symptom for orgs/users whose login casing isn't perfectly consistent between the stored installation row and the request path.

No test in `apps/api/tests/test_token_resolution.py` currently exercises a case-differing `account_login`, so this could regress silently again.

Normalize `account_login` casing consistently at write time (e.g. always store lowercased) and/or make `get_for_org`/`get_for_user` compare case-insensitively (e.g. `func.lower(GitHubInstallation.account_login) == account_login.lower()`), matching the pattern already used by `assert_owner_matches_org` and the installation-verification checks.

nazarli-shabnam:
## Summary

Closes #246.

`installation_repo.get_for_org`/`get_for_user` matched `account_login` with an exact `==` comparison. `account_login` is stored verbatim from GitHub's install payload (`installations.py`), but RBAC/ownership checks elsewhere in the same request path already compare logins case-insensitively (`assert_owner_matches_org` in `core/rbac.py`, `_verify_installation` in `installations.py`).

## Failure scenario

If a caller reaches a repo/analytics/automation endpoint with an `owner`/`org_login` path parameter whose casing differs from what's stored (a GitHub org renamed casing, or a client simply sending different case), the RBAC/ownership check passes (it's case-insensitive) but `token_resolution.resolve_org_token`/`resolve_personal_token`'s installation lookup fails to find a row that's actually there — silently falling back to a client-supplied PAT, or raising `NoGitHubTokenAvailable` even though a valid installation exists.

## What changed

- `get_for_org` and `get_for_user` now compare `account_login` via `func.lower(...) == account_login.lower()`, matching the case-insensitive comparisons used elsewhere in this same request path.
- `upsert`'s own lookup (used at install-sync time) is intentionally left as an exact match — that's a narrower, separate question about how install-sync dedupes rows on casing drift, not what this issue is about; changing it wasn't needed to close the gap between RBAC checks and token resolution.
- Added `test_get_for_org_matches_regardless_of_login_casing` and `test_get_for_user_matches_regardless_of_login_casing` to `apps/api/tests/test_installation_repo.py`.

## Why

Directly relevant to the just-closed "GitHub org not connecting" bug class on this branch — this is a related edge case the recent fix didn't cover, and the RBAC/token-resolution mismatch it produces (installation exists in the DB but token resolution can't see it) is exactly the "silently falls back or fails oddly" pattern that class of bug was about.

**This PR touches token-resolution code**, flagging per this repo's guardrails: the change is a comparison-only fix (case-insensitive match instead of exact match) to an existing, already-tested lookup function — no change to what gets returned once a row is found, no change to encryption/minting logic.

No migrations, no RBAC bypass — purely a lookup-matching fix.

## Test plan
- [x] `pytest tests/test_installation_repo.py -v` — all 6 tests pass (2 new)
- [x] `pytest tests/test_token_resolution.py tests/test_installations.py -v` — all 42 tests pass, no regressions
- [ ] CI

### DIFF
--- a/apps/api/src/repositories/installation_repo.py
+++ b/apps/api/src/repositories/installation_repo.py
@@ -1,3 +1,4 @@
+from sqlalchemy import func
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session
 
@@ -84,9 +85,16 @@ def create(
 
 
 def get_for_org(db: Session, org_id: int, account_login: str) -> GitHubInstallation | None:
+    # Case-insensitive: account_login is stored verbatim from GitHub's install payload,
+    # but RBAC/ownership checks elsewhere (assert_owner_matches_org, _verify_installation)
+    # already compare logins case-insensitively -- an exact match here could pass those
+    # checks yet fail to find an installation that's actually there (#246).
     return (
         db.query(GitHubInstallation)
-        .filter(GitHubInstallation.org_id == org_id, GitHubInstallation.account_login == account_login)
+        .filter(
+            GitHubInstallation.org_id == org_id,
+            func.lower(GitHubInstallation.account_login) == account_login.lower(),
+        )
         .first()
     )
 
@@ -96,7 +104,7 @@ def get_for_user(db: Session, owner_user_id: int, account_login: str) -> GitHubI
         db.query(GitHubInstallation)
         .filter(
             GitHubInstallation.owner_user_id == owner_user_id,
-            GitHubInstallation.account_login == account_login,
+            func.lower(GitHubInstallation.account_login) == account_login.lower(),
         )
         .first()
     )
--- a/apps/api/tests/test_installation_repo.py
+++ b/apps/api/tests/test_installation_repo.py
@@ -77,3 +77,42 @@ def test_upsert_reraises_when_the_integrity_error_was_not_actually_a_race(db):
             installation_id=1,
             org_id=999999,
         )
+
+
+def test_get_for_org_matches_regardless_of_login_casing(db):
+    # account_login is stored verbatim from GitHub's install payload; RBAC/ownership checks
+    # elsewhere (assert_owner_matches_org, _verify_installation) already compare logins
+    # case-insensitively, so this lookup must too -- otherwise a case mismatch (e.g. a
+    # renamed org, or a caller passing different casing) passes those checks but fails to
+    # find an installation that's actually there (#246).
+    org_id = _acme_org_id(db)
+    installation_repo.upsert(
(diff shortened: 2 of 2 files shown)


nazarli-shabnam:
Confirmed — get_for_org/get_for_user now compare account_login case-insensitively via func.lower(), matching the RBAC checks that already do this elsewhere. Tests cover both lookups. CI green. Thanks!

PR #258, opened 2026-07-24
fix(api): unbounded in-process caches never evict expired entries

### LINKED ISSUE #247: Unbounded in-process response caches never evict entries
## Summary

Two module-level in-process caches never remove expired or stale entries — they only ever grow:

- `_stats_cache` — `apps/api/src/routers/repos.py:28`: `dict[tuple[str, str, str], tuple[float, RepoStatsResponse]]`, keyed by `(owner, repo, token_hash)`.
- `_events_cache` — `apps/api/src/routers/github.py:51`: keyed by `(org, token_hash, per_page)`.

Both use a TTL check (`_STATS_CACHE_TTL_SECONDS` / `_EVENTS_CACHE_TTL_SECONDS`) to decide whether a cached entry is *stale*, but a stale entry is simply overwritten on next access — nothing ever deletes an entry that's no longer being requested. There is no size cap, no LRU eviction, and no periodic sweep.

## Failure scenario

GitHub App installation tokens (the preferred auth path per `token_resolution.py`) are minted per-request and rotate roughly hourly. Since `token_hash` is part of the cache key (`hashlib.sha256(token.encode()).hexdigest()`), every token rotation for every org/repo combination that's ever been queried adds a *new* cache entry rather than replacing an old one. Over the lifetime of a long-running API process serving many orgs and repos, `_stats_cache` and `_events_cache` grow without bound — steady memory growth that's only resolved by a process restart.

## Why this matters

This is a slow-burn production reliability issue: a self-hosted instance running for weeks against many connected orgs will accumulate cache memory it never releases, potentially leading to OOM under sustained multi-tenant usage. It's the same category of concern already called out for `_get_all_pages`'s `_MAX_PAGES` cap in `packages/checks/src/checks/runner.py` (issue #214) — this is the analogous gap on the caching side rather than the pagination side.

Add eviction of expired entries (e.g. sweep on write, or switch to a bounded LRU/TTL cache structure) so the cache size stays proportional to *currently relevant* org/repo/token combinations rather than every combination ever seen.

nazarli-shabnam:
## Summary

Closes #247.

`apps/api/src/routers/repos.py`'s `_stats_cache` and `apps/api/src/routers/github.py`'s `_events_cache` are module-level dicts keyed by `(owner, repo, token_hash)` / `(org, token_hash, per_page)` with TTL-based staleness checks — but expired entries were only ever *skipped* on a cache-hit check, never actually removed from the dict.

Installation tokens rotate hourly, changing `token_hash`, so every org × repo × hour of uptime added a new entry that was never pruned. A long-running instance serving many orgs sees steady, unbounded memory growth in these two dicts.

## What changed

- `repos.py`: added `_evict_expired_stats(now)`, called right after inserting a fresh entry on every cache miss — sweeps any entry whose TTL has already elapsed.
- `github.py`: same pattern, `_evict_expired_events(now)`.
- Added a test in each file's test suite (`test_repo_stats_evicts_expired_entries_from_the_cache`, `test_events_evicts_expired_entries_from_the_cache`) that seeds a synthetic stale entry, triggers a real cache miss, and asserts the stale entry is gone afterward.

## Why

Both caches are process-lifetime dicts with no other cleanup path — an operator running Clevis for weeks/months against a growing set of orgs would otherwise see these grow forever. The fix piggybacks the sweep on the existing miss path rather than adding a background thread/scheduler, since a miss already happens on roughly the same cadence as new stale entries would accumulate.

No RBAC/auth/token-encryption/migration changes — purely an in-memory cache-hygiene fix, no behavior change to cache hits/misses themselves.

## Test plan
- [x] `pytest tests/test_repos.py tests/test_github_events.py -v` — all 40 tests pass (2 new)
- [ ] CI

### DIFF
--- a/apps/api/src/routers/github.py
+++ b/apps/api/src/routers/github.py
@@ -124,6 +124,14 @@ def _fetch_events(org_login: str, token: str, per_page: int) -> OrgEventsRespons
     return OrgEventsResponse(org=org_login, events=events)
 
 
+def _evict_expired_events(now: float) -> None:
+    # Same reasoning as repos.py's _evict_expired_stats: token_hash rotates hourly with
+    # each fresh installation token, so stale keys would otherwise accumulate forever.
+    expired = [key for key, (cached_at, _) in _events_cache.items() if now - cached_at >= _EVENTS_CACHE_TTL_SECONDS]
+    for key in expired:
+        del _events_cache[key]
+
+
 def _cached_events(org_login: str, token: str, per_page: int) -> OrgEventsResponse:
     key = (org_login, _token_hash(token), per_page)
     now = time.monotonic()
@@ -132,6 +140,7 @@ def _cached_events(org_login: str, token: str, per_page: int) -> OrgEventsRespon
         return cached[1]
     events = _fetch_events(org_login, token, per_page)
     _events_cache[key] = (now, events)
+    _evict_expired_events(now)
     return events
 
 
--- a/apps/api/src/routers/repos.py
+++ b/apps/api/src/routers/repos.py
@@ -105,6 +105,15 @@ def _fetch_stats(owner: str, repo: str, token: str) -> RepoStatsResponse:
     }
 
 
+def _evict_expired_stats(now: float) -> None:
+    # Installation tokens rotate hourly, so every org x repo x hour of uptime adds a new
+    # key that's never revisited once its token_hash goes stale -- without this sweep the
+    # dict grows without bound for a long-running instance serving many orgs.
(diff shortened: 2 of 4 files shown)


nazarli-shabnam:
Confirmed — both stats and events caches now evict expired entries on write, closing the unbounded-growth issue in #247. Tests reproduce the stale-entry cleanup directly. CI green. Thanks!

nazarli-shabnam (commented):
Confirmed — both stats and events caches now evict expired entries on write, closing the unbounded-growth issue in https://github.com/nazarli-shabnam/clevis/issues/247. Tests reproduce the stale-entry cleanup directly. CI green.

PR #259, opened 2026-07-24
fix(api): duplicate pending invitations aren't prevented

### LINKED ISSUE #248: Duplicate pending invitations aren't prevented
## Summary

`create_invitation` (`apps/api/src/routers/invitations.py:52-63`) unconditionally calls `invitation_repo.create` with no check for an existing pending invite to the same email in the same org:

```python
@router.post("/orgs/{org_login}/invitations", response_model=InvitationCreateResponse)
def create_invitation(
    body: InvitationCreate,
    ctx: OrgContext = Depends(require_org_role(min_role="admin")),
    user: UserOut = Depends(require_auth),
    db: Session = Depends(get_db),
):
    invitation = invitation_repo.create(db, org_id=ctx.org.id, email=body.email, invited_by_user_id=user.id)
    ...
```

## Failure scenario

An org admin calls `POST /orgs/{org_login}/invitations` twice for the same email (double-click, accidental resend, a retried request after a slow response) — this creates two separate `invitations` rows, each with its own token, each independently valid and independently acceptable via `POST /invitations/{token}/accept`.

`revoke_invitation` (same file, lines 86-100) only revokes the specific invitation ID passed in. If the admin later decides to revoke access for that email (e.g. they invited the wrong person, or the person left before accepting), the UI most likely surfaces only the latest invitation to revoke — the older, forgotten invitation is still `pending` and can still be used to accept and join the org.

## Why this matters

This is a minor access-control gap: an admin's belief that "I revoked bob@x.com's invite" doesn't hold if a stale duplicate invite is still floating around. It's low-severity (both invites still require the invited email to match and be verified, per issue #217's flow) but easy to hit by accident and easy to fix.

Before creating a new invitation, check for an existing `pending` (non-expired) invitation to the same `(org_id, email)` pair and either reuse/refresh it (update `expires_at`, reuse the row) or reject the duplicate create with a clear error, rather than silently allowing multiple simultaneously-valid invites.

nazarli-shabnam:
## Summary

Closes #248.

`POST /orgs/{org_login}/invitations` had no check for an existing pending invite to the same email within the same org. Inviting the same address twice (by accident, or a retried request) created two independently valid, independently tokened pending invitations.

## Failure scenario

An admin invites `bob@acme.com`, then invites them again later (forgetting the first invite, or after a client-side retry). Two separate `Invitation` rows now exist, each with its own token and its own link. Revoking "the" invite shown in the admin UI (most likely the most recently created one) leaves the older link still valid and able to grant org membership until it separately expires or is separately found and revoked — a stale invite link staying alive longer than the admin believes it does.

## What changed

- Added `invitation_repo.get_pending_for_org_and_email(db, org_id, email)` — same `Invitation.email.ilike(...)` case-insensitive pattern `list_pending_for_email` already uses for the (cross-org) registration-time lookup, but scoped to a single org and filtered to `status == "pending"` and not-yet-expired.
- `create_invitation` now checks this before creating a new row and returns `409 Conflict` if a pending invite already exists for that org/email.
- A new invite can still be sent once the existing one is revoked or naturally expires — this only blocks a duplicate *pending* invite, not re-inviting after the old one is gone.
- Added tests: rejects a duplicate while pending, case-insensitive match, and that a new invite is allowed after the prior one is revoked or expires.

## Why

Directly closes the gap the issue describes — no schema/behavior change beyond adding the pre-create check; existing single-invite flows are unaffected.

No RBAC/auth/token-encryption/migration changes.

## Test plan
- [x] `pytest tests/test_invitations.py -v` — all 18 tests pass (4 new)
- [ ] CI

### DIFF
--- a/apps/api/src/repositories/invitation_repo.py
+++ b/apps/api/src/repositories/invitation_repo.py
@@ -39,6 +39,19 @@ def list_for_org(db: Session, org_id: int) -> list[Invitation]:
     return db.query(Invitation).filter(Invitation.org_id == org_id).order_by(Invitation.created_at.desc()).all()
 
 
+def get_pending_for_org_and_email(db: Session, org_id: int, email: str) -> Invitation | None:
+    return (
+        db.query(Invitation)
+        .filter(
+            Invitation.org_id == org_id,
+            Invitation.email.ilike(email),
+            Invitation.status == "pending",
+            Invitation.expires_at > datetime.now(timezone.utc),
+        )
+        .first()
+    )
+
+
 def list_pending_for_email(db: Session, email: str) -> list[Invitation]:
     return (
         db.query(Invitation)
--- a/apps/api/src/routers/invitations.py
+++ b/apps/api/src/routers/invitations.py
@@ -56,6 +56,12 @@ def create_invitation(
     user: UserOut = Depends(require_auth),
     db: Session = Depends(get_db),
 ):
+    existing = invitation_repo.get_pending_for_org_and_email(db, org_id=ctx.org.id, email=body.email)
+    if existing is not None:
+        raise HTTPException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail=f"A pending invitation already exists for {body.email} in this organization",
+        )
     invitation = invitation_repo.create(db, org_id=ctx.org.id, email=body.email, invited_by_user_id=user.id)
     return {
         "invitation": invitation,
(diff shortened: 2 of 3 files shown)


nazarli-shabnam (commented):
Confirmed — create_invitation now rejects a duplicate pending, non-expired invite for the same org+email (case-insensitive), matching what https://github.com/nazarli-shabnam/clevis/issues/248 asked for, with solid test coverage including the revoked/expired re-invite cases. CI green.

PR #261, opened 2026-07-24
fix(api): distinguish a mint failure from no installation at all

### LINKED ISSUE #250: Installation-token-mint failures produce a misleading 'not installed' error message
## Summary

`_from_installation` (`apps/api/src/services/token_resolution.py:29-45`) catches token-minting failures and returns `None`, causing the caller to silently fall back to a client-supplied PAT (or raise `NoGitHubTokenAvailable` if none is supplied):

```python
except (httpx.HTTPStatusError, httpx.RequestError):
    # The installation row is stale (App uninstalled/suspended on GitHub's side) or
    # GitHub's token-minting endpoint is briefly unreachable. Don't let this become an
    # unhandled 500 — fall back to a client-supplied token same as "no installation".
    logger.warning(...)
    return None
```

`resolve_org_token`/`resolve_personal_token` (same file, lines 52-83) then raise, when no PAT is available:

```python
raise NoGitHubTokenAvailable(
    f"No GitHub App installation found for '{account_login}' and no token was provided. "
    "Install the GitHub App for this organization, or add a personal access token in Settings."
)
```

## Failure scenario

An org's `github_installations` row exists and is valid in the local DB, but the installation was uninstalled or suspended on GitHub's side (e.g. an org admin removed the app from GitHub directly, without Clevis's webhook handler processing the uninstall event yet, or the App itself was suspended). A user hits an endpoint needing that org's token, gets `_from_installation` returning `None` after a caught `HTTPStatusError`/`RequestError`, and — with no PAT configured — sees: *"No GitHub App installation found for 'acme'... Install the GitHub App..."* This message is factually wrong: the installation **is** found in the DB; it's just failing to mint a token. Anyone troubleshooting this (an operator or the org admin) would go try to "install" something that already shows as installed in the Clevis UI, instead of investigating why token minting is failing (suspended app, revoked permissions, GitHub API outage, etc).

## Why this matters

Misleading error messages waste real troubleshooting time, especially for self-hosted operators without direct access to server logs (the actual root cause — the `HTTPStatusError`/`RequestError` — is only visible in the `logger.warning(..., exc_info=True)` line, not surfaced to the user/API response at all).

Distinguish "no installation row exists" from "installation row exists but token minting failed" in the error surfaced to the caller — e.g. a distinct exception/message for the latter case that hints at "installation may need to be reconnected" rather than "install the GitHub App."

nazarli-shabnam:
## Summary

Closes #250.

`token_resolution._from_installation` silently falls back to `None` (then to a client-supplied token, or `NoGitHubTokenAvailable` if there's none of those either) both when there's genuinely no installation row *and* when an installation exists but minting a token for it failed (stale/uninstalled on GitHub's side, or a transient API error — see the warning it already logs). `resolve_org_token`/`resolve_personal_token` couldn't tell these two cases apart, so whoever hits the final error always saw the same generic "No GitHub App installation found... Install the GitHub App..." message — even when the installation row genuinely exists in the DB and just needs reinstalling or the App needs reconfiguring.

## What changed

- Added `_no_token_error(account_login, installation_exists, *, personal)` in `token_resolution.py`, which builds a distinct message when `installation_exists` is `True`: *"A GitHub App installation exists for '{account_login}' but minting a token for it failed. The installation may need to be reinstalled, or the GitHub App may be misconfigured -- check server logs for details."*
- `resolve_org_token`/`resolve_personal_token` now pass whether an installation row was actually found (not just whether minting succeeded) into the error builder.
- Added tests asserting the mint-failure message is distinct from ("does not contain") the "Install the GitHub App" wording, for both the org and personal resolvers.

## Why

Directly closes the gap the issue describes: someone debugging a broken installation would previously be told to "install" something that, per the DB, is already installed — sending them down the wrong path.

**This PR touches token-resolution code**, flagging per this repo's guardrails: purely an error-message/branching change based on data already being looked up (`installation` was already fetched before this change) — no change to what gets minted, returned, or persisted.

No migrations, no RBAC bypass, no behavior change to the success paths.

## Test plan
- [x] `pytest tests/test_token_resolution.py -v` — all 10 tests pass (2 new)
- [x] `pytest tests/test_analytics_cockpit.py tests/test_analytics_my_view.py tests/test_collab.py tests/test_repos.py tests/test_github_events.py -q` — all 110 pass, no regressions (none of these assert on the exact error text)
- [ ] CI

### DIFF
--- a/apps/api/src/services/token_resolution.py
+++ b/apps/api/src/services/token_resolution.py
@@ -49,6 +49,24 @@ def _github_app_configured() -> bool:
     return settings.github_app_id is not None and settings.github_app_private_key is not None
 
 
+def _no_token_error(account_login: str, installation_exists: bool, *, personal: bool) -> NoGitHubTokenAvailable:
+    if installation_exists:
+        # Distinct from "never installed" -- the installation row is there, minting just
+        # failed (stale/uninstalled on GitHub's side, or a transient API error; see the
+        # warning _from_installation already logged). Telling the caller to "install" it
+        # sends them down the wrong path when, per the DB, it's already installed.
+        return NoGitHubTokenAvailable(
+            f"A GitHub App installation exists for '{account_login}' but minting a token for it "
+            "failed. The installation may need to be reinstalled, or the GitHub App may be "
+            "misconfigured -- check server logs for details."
+        )
+    install_hint = "Install the GitHub App" if personal else "Install the GitHub App for this organization"
+    return NoGitHubTokenAvailable(
+        f"No GitHub App installation found for '{account_login}' and no token was provided. "
+        f"{install_hint}, or add a personal access token in Settings."
+    )
+
+
 def resolve_org_token(db: Session, *, org_id: int, account_login: str, client_token: str | None) -> str:
     installation = (
         installation_repo.get_for_org(db, org_id=org_id, account_login=account_login)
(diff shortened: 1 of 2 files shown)


nazarli-shabnam (commented):
Confirmed — token resolution now distinguishes a mint failure on an existing installation from no installation at all, exactly as https://github.com/nazarli-shabnam/clevis/issues/250 asked, using data already fetched at both call sites. Tests cover both branches. CI green.

PR #262, opened 2026-07-24
fix(worker): fence terminal job writes against a reclaim-and-reclaim race

### LINKED ISSUE #253: Narrow race between stale-job reclaim and a still-alive job handler
## Summary

`_mark_done` / `_mark_failed` in `apps/worker/src/worker.py` (around lines 203-239) guard their terminal status update with `WHERE status='processing'`, relying on the job still being in `processing` state at the moment the handler finishes. Separately, `_reclaim_stale_jobs` (lines 87-106) periodically resets jobs whose `heartbeat_at` has gone stale (past `RECLAIM_TIMEOUT_MINUTES`) back to `queued` (or `failed` if retries are exhausted), on the assumption that a stale heartbeat means the worker processing it has crashed.

## Failure scenario

The `_JobHeartbeat` background thread (which touches `heartbeat_at` every ~10s while a handler runs, per its design intent — see AGENTS.md's description of issue #215) can, in principle, die independently of the main thread that's actually still executing the GitHub API call (e.g. an unexpected exception in the heartbeat thread itself, or it silently stops without crashing the process). If that happens and enough time passes for `_reclaim_stale_jobs` to sweep the job back to `queued`/`failed` while the *original* handler is still alive and about to finish its in-flight GitHub call (within its own request timeout), two things can happen almost simultaneously:

1. The original handler finishes and calls `_mark_done`/`_mark_failed`, but the `WHERE status='processing'` guard no longer matches (the reclaim already flipped it to `queued`/`failed`) — the update becomes a silent no-op, and the original handler's outcome is lost.
2. A second worker replica picks up the now-`queued` job via the normal `SELECT ... FOR UPDATE SKIP LOCKED` claim and starts processing it again — concurrently with the first handler's still-in-flight GitHub call.

For a job like clearing an Actions cache, this means two workers could call GitHub's cache-clear endpoint for the same job at nearly the same time.

This is explicitly a known, documented trade-off in the current design (the heartbeat mechanism exists specifically to distinguish a slow-but-alive job from a crashed one, per issue #215), and requires the heartbeat thread to die independently of the main thread — a low-probability scenario. Filing as a tracked issue rather than leaving it as an implicit gap, since the failure mode (double-processing a GitHub-mutating job) is worth a deliberate decision rather than an accepted silent risk.

## Why this matters

Low probability, but the blast radius (duplicate external side effects against the GitHub API) is more significant than the average bug — worth a conscious decision (accept as-is, add idempotency, or add a stronger guard) rather than leaving it undocumented as a GitHub issue.

Options to consider (not prescribing one): make the job handler itself idempotent so double-processing is harmless; have `_mark_done`/`_mark_failed` check `retry_count`/a monotonic "claim token" rather than only `status='processing'`, so a genuinely-still-alive handler can detect it was reclaimed and abort cleanly instead of racing; or tighten the heartbeat thread's own failure handling so it can never silently stop without also stopping the main handler.

nazarli-shabnam:
## Summary

Closes #253.

`_mark_done`/`_mark_failed` guard their terminal update with `WHERE status='processing'`. This protects the simple case (reclaim sweep reset the job back to `'queued'`/`'failed'` -- status no longer matches, safe no-op). It doesn't protect a narrower race: the reclaim sweep resets a stale job to `'queued'` (bumping `retry_count`) **and** a second worker's own `SELECT ... FOR UPDATE SKIP LOCKED` picks it back up before the first (stale, still in-flight) worker's terminal write runs. By then `status` is `'processing'` again -- from the *second* worker's claim -- so the guard alone incorrectly matches, letting the first worker's stale completion clobber the second worker's actively-processing row (mislabeling it done/failed while the second worker is still mid-flight, or double-hitting GitHub's cache-clear endpoint for the same job).

## What changed

- `_mark_done`, `_mark_failed`, and `_requeue_for_retry` now also require `retry_count` to equal the value this worker observed when it originally claimed the job (already available everywhere via the existing `retry_count` parameter threaded through `process_job`/handlers). The reclaim sweep always increments `retry_count` when it resets a job, so a second worker's re-claim carries a `retry_count` the first worker's stale write no longer matches -- the update becomes a no-op instead of clobbering the second worker's row.
- No schema change: `retry_count` already existed specifically as a fencing/dedup signal, it just wasn't checked on the terminal-write side, only used to compute the next retry count.
- Updated the existing `_requeue_for_retry`-touching tests for the new 4th SQL param, and added `test_process_job_terminal_write_is_a_noop_if_a_second_worker_reclaimed_and_reprocessed_it`, which reproduces the actual scenario the issue describes (reclaim + re-claim, not just reclaim alone) and asserts the second worker's `processing` row is left untouched.

## Why

Directly closes the narrow window the issue flagged as an accepted trade-off in the original heartbeat design (#215) -- worth a tracked, deliberate fix given how sensitive double-processing a privileged job (clearing GitHub Actions caches) is, even though it's low-probability in practice.

No migrations, no RBAC bypass, no behavior change to the normal (non-racing) success/failure/retry paths.

## Test plan
- [x] `pytest apps/worker/tests/ -v` — all 35 tests pass (1 new, 3 updated for the new param)
- [ ] CI

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale job results from overwriting work reclaimed by another worker.
  * Improved job retry and completion handling by validating the latest retry state.
  * Jobs with invalid tokens or unexpected errors are now marked failed with a sanitized error message.
  * GitHub action failures now correctly record non-success responses and preserve job state integrity.

### DIFF
--- a/apps/worker/src/worker.py
+++ b/apps/worker/src/worker.py
@@ -84,43 +84,53 @@ class ClearActionsCachePayload(BaseModel):
     ref: str | None = None
 
 
-def _mark_done(conn: psycopg.Connection, job_id: int, result: dict) -> None:
-    # WHERE status='processing' guards against a lost update if the reclaim sweep
-    # already reset this job (e.g. this worker stalled past RECLAIM_TIMEOUT_MINUTES)
-    # out from under us — the write becomes a safe no-op instead of clobbering
-    # whatever state/retry_count the reclaim (or another worker) since wrote.
+def _mark_done(conn: psycopg.Connection, job_id: int, result: dict, expected_retry_count: int) -> None:
+    # WHERE status='processing' AND retry_count=expected_retry_count fences this update
+    # against not just a lost update (reclaim already reset this job out from under us --
+    # status no longer 'processing') but also the narrower race where a *second* worker
+    # has since re-claimed the same job: reclaim bumps retry_count when it resets a stale
+    # job back to 'queued', so if that happened and another worker's SELECT ... FOR UPDATE
+    # picked it up again, status is back to 'processing' but retry_count no longer matches
+    # what *this* worker observed when it originally claimed the job. Without the
+    # retry_count check, this stale completion would silently clobber the second worker's
+    # in-flight row (issue #253).
     with conn.cursor() as cur:
         cur.execute(
-            "UPDATE jobs SET status='done', result=%s, updated_at=NOW() WHERE id=%s AND status='processing'",
-            (json.dumps(result), job_id),
+            "UPDATE jobs SET status='done', result=%s, updated_at=NOW() "
+            "WHERE id=%s AND status='processing' AND retry_count=%s",
+            (json.dumps(result), job_id, expected_retry_count),
         )
     conn.commit()
 
 
-def _mark_failed(conn: psycopg.Connection, job_id: int, error_text: str) -> None:
+def _mark_failed(conn: psycopg.Connection, job_id: int, error_text: str, expected_retry_count: int) -> None:
     with conn.cursor() as cur:
         cur.execute(
-            "UPDATE jobs SET status='failed', result=%s, updated_at=NOW() WHERE id=%s AND status='processing'",
-            (error_text, job_id),
+            "UPDATE jobs SET status='failed', result=%s, updated_at=NOW() "
+            "WHERE id=%s AND status='processing' AND retry_count=%s",
+            (error_text, job_id, expected_retry_count),
         )
     conn.commit()
 
 
 def _requeue_for_retry(conn: psycopg.Connection, job_id: int, retry_count: int, error_text: str) -> None:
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
Logic is correct — fencing _mark_done/_mark_failed/_requeue_for_retry on retry_count in addition to status closes the two-worker reclaim race from #253, and the new test reproduces the actual interleaving. CI's Python Tests + Compile was failing though: the retry_count fencing change touched the token-decrypt-failure branch's _mark_failed call, but nothing exercised that path, so diff-cover failed on the changed line. Pushed a test covering it (50172b4) — CI is green now. Thanks!

nazarli-shabnam:
Fixed the CodeRabbit finding — the decrypt-failure test now asserts the retry_count=%s predicate and its parameter value, not just status/job id (e32ddfc). CI still green.

PR #263, opened 2026-07-24
fix(checks): security checks misclassify HTTP errors, hiding real results

### LINKED ISSUE #245: Security checks misclassify HTTP errors, silently hiding real pass/fail results
## Summary

Five related bugs in `packages/checks/src/checks/github_checks.py` and `packages/checks/src/checks/runner.py`, all stemming from how HTTP error codes are classified during a scan. Net effect: transient GitHub-side hiccups (5xx, rate limits) get reported as *real security failures*, and partial access-scope problems get reported as *all clear* — the opposite of what each check's own comments say it's trying to guarantee.

## Bugs

**1. `DefaultBranchNoForcePushCheck.run` treats a fully-unprotected branch as "unknown" instead of "force push allowed"**
`packages/checks/src/checks/github_checks.py:313-322`

A 404 from `GET /repos/{owner}/{repo}/branches/{branch}` means the branch has **no protection at all** — force pushes are unambiguously allowed. But the handler is a blanket `except httpx.HTTPError`, which buckets 404 together with 403/429/network errors as "unknown" and excludes the repo from `checked` entirely:

```python
try:
    details = _get(f"{base_url}/repos/{owner}/{repo['name']}/branches/{branch}", token)
except httpx.HTTPError:
    unknown += 1
    continue
```

Failure scenario: an org where every repo's default branch is completely unprotected (the exact worst case this check exists to catch) ends up with `checked == 0` → `status: "error"`, never `"fail"`.

**2. `_branch_protection_status` classifies unexpected error codes as "unprotected" instead of "unknown"**
`packages/checks/src/checks/github_checks.py:90-96`

```python
def _branch_protection_status(exc):
    code = exc.response.status_code
    if code == 404: return "unprotected"
    if code in (403, 429): return "unknown"
    return "unprotected"   # <-- also catches 500/502/503/etc.
```

`BranchProtectionEnabled.run` (lines 146-165) then counts that repo as checked-but-not-protected. Failure scenario: a transient 500/502 from GitHub while checking one repo's branch protection can flip the entire org's branch-protection check to `"fail"`, even though every real repo is protected.

**3. `DependabotAlertsCheck` / `CodeScanningCheck` report "pass" when most repos are forbidden**
`packages/checks/src/checks/github_checks.py:238-241` and `286-288`

Both only escalate to `status: "error"` when **every** repo in the org returns 403:

```python
if forbidden == len(repos):
    return {"status": "error", "value": counts}
compliant = counts["critical"] == 0 and counts["high"] == 0
return {"status": "pass" if compliant else "fail", "value": counts}
```

Failure scenario: 10 repos, 9 return 403 (token missing `security_events` scope), 1 accessible repo has 0 critical/high alerts → the check reports `{"status": "pass", ...}`, even though the alert status for 9 of 10 repos is completely unknown. This directly contradicts the code's own comment a few lines above ("403 means unknown, not zero, so it must not silently count toward clear") — the counts themselves are correct, but the final pass/fail verdict is computed as if the forbidden repos didn't exist.

**4. `_get_with_retry` never retries GitHub 5xx errors**
`packages/checks/src/checks/runner.py:29-44`

The retry loop only retries on `httpx.RequestError` (connection-level), a `429`, or a rate-limit-flavored `403`. A `500/502/503` is returned as-is on the first attempt, and `_get()` immediately calls `raise_for_status()`, raising `HTTPStatusError` with no retry. This is what feeds bug #2 above — a single flaky 500 gets treated as a hard failure instead of being retried.

**5. Rate-limit backoff ignores the actual `Retry-After` value**
`packages/checks/src/checks/runner.py:19-26, 40-42`

`_is_secondary_rate_limit` detects the *presence* of a `Retry-After` header but the caller never reads its value — it always sleeps a fixed `2**attempt` (1s, then 2s) regardless of what GitHub actually asked for. If GitHub sends `Retry-After: 60`, the check retries after 1-2s anyway, likely re-hits the same secondary rate limit, and only gets 3 attempts total before giving up and raising (surfacing as a per-check `"error"` in `run_all_checks`).

## Why this matters

These checks drive the security score shown to org admins and the score-trend chart (`scan_results` table). A scan that happens to hit a transient GitHub error can make a compliant org look non-compliant (bugs #1, #2, #4), while a scan with a partial permission problem can make a genuinely non-compliant org look clean (bug #3). Both directions undermine the tool's core value proposition.

- Distinguish "genuinely unprotected" (404) from "couldn't evaluate" (403/429/5xx/network) consistently across all three checks that hit the branches endpoint, and make sure "unprotected" always counts toward `checked`/fail, never toward `unknown`.
- Change the Dependabot/CodeScanning "error" threshold from `forbidden == total` to something that reflects partial unknowns (e.g. don't report "pass" unless the whole repo set was evaluable, or surface an explicit "partial" status).
- Extend `_get_with_retry` to retry on 5xx (with backoff), and read/respect the actual `Retry-After` header value when present instead of a fixed exponential schedule.

No fix included in this issue — filed for tracking after a project-wide bug sweep; happy to pick this up if wanted.

nazarli-shabnam:
## Summary

Closes #245.

Five related bugs in `packages/checks/src/checks/github_checks.py`, all stemming from how the retry loop and per-check exception handlers treat non-2xx responses. Every real-world effect is the same shape: transient GitHub-side hiccups (5xx, rate limits) got reported as real security failures, and partial access-scope problems got reported as "all clear" -- the opposite of what each check's own existing comments say it's trying to guarantee.

## What changed

1. **`DefaultBranchNoForcePushCheck`** -- a 404 from `/repos/{owner}/{repo}/branches/{branch}` means the default branch has **no protection at all**, so force pushes are unambiguously allowed -- not "unknown". It was bucketed with the genuine unknowns (403/429/network), so an org where every repo's default branch was completely unprotected (the exact worst case this check exists to catch) reported `checked == 0` -> `"error"`, never `"fail"`. Now a 404 counts toward `checked` **and** `force_push_allowed`; 403/429/network errors still correctly stay excluded from the denominator.

2. **`_branch_protection_status`** -- fell through to `"unprotected"` for any status code that wasn't 404/403/429 (a transient 500/502, most notably). A flaky GitHub 5xx during a scan could flip an org's entire branch-protection check to `"fail"` even though every real repo was protected. Now anything but 404 maps to `"unknown"`.

3. **`DependabotAlertsCheck` / `CodeScanningCheck`** -- only escalated to `"error"` when *every* repo 403'd. If 9 of 10 repos are forbidden (e.g. a fine-grained PAT missing a scope) and 1 is clean, the check reported `"pass"` based solely on the one reachable repo -- contradicting the code's own comment ("403 means unknown, not zero, so it must not silently count toward clear"). Now a would-be `"pass"` is downgraded to `"error"` whenever any repo was forbidden; a `"fail"` from real findings on the reachable repos stays valid regardless of unknowns elsewhere (a real problem doesn't need corroboration from the inaccessible repos to be real).

4. **`_get_with_retry`** -- only retried on `httpx.RequestError`, a `429`, or a rate-limit-flavored `403`. A `500/502/503` was returned as-is and `_get()` immediately raised via `raise_for_status()` -- no retry, directly feeding bug #2 above. Now retries a 5xx too, presumed transient same as every other GitHub-talking client in this codebase (`apps/api/src/services/github_client.py`).

5. **Retry-After handling** -- the loop detected the header's *presence* but never read its value, always sleeping a fixed `2**attempt` (1s, then 2s). If GitHub asks for a 60s backoff, the loop burned its 3 attempts hitting the same limit again instead of eventually succeeding. Now honors the actual value (capped at 60s as a defensive ceiling against a malformed/huge value) when present, falling back to exponential backoff otherwise.

## A parallel issue not fixed here

`apps/api/src/services/github_client.py` has the same `_is_secondary_rate_limit`/retry-loop pattern (doesn't retry 5xx, ignores Retry-After's value either) -- out of scope for #245, which only covers `packages/checks`, so left untouched here. Worth a separate follow-up if desired.

## What changed in tests

- Replaced `test_dependabot_mixed_403_and_clean_repos_still_passes` and added a `CodeScanningCheck` equivalent -- both directly asserted the old buggy behavior (mixed 403 + otherwise-clean -> `"pass"`). Now split into two tests each: the false-clean case (now `"error"`) and a real-alert-plus-403 case (stays `"fail"`).
- Replaced `test_force_push_unprotected_branch_excluded_from_denominator` (asserted the old bug: 404 -> excluded, `checked == 0`) with `test_force_push_unprotected_branch_404_means_force_push_is_allowed` (404 -> `checked=1, force_push_allowed=1, status="fail"`), plus a new test for the still-correctly-excluded 403/429 case.
- Added `test_branch_protection_5xx_counts_as_unknown_not_unprotected`, `test_get_retries_on_5xx_then_succeeds`, `test_get_honors_the_actual_retry_after_value_not_a_blind_backoff`, `test_get_caps_an_excessive_retry_after_value`.

No RBAC/auth/token-encryption/migration changes -- this is a `packages/checks` library fix only, consumed by `apps/api`'s analytics/security-matrix routes via the editable install; ran the affected `apps/api` test files as a regression check (all pass, none assert on internals this touches).

## Test plan
- [x] `pytest packages/checks/tests/ -v` -- all 43 tests pass (8 new/replaced)
- [x] `pytest apps/api/tests/test_analytics_scoring.py apps/api/tests/test_repos.py apps/api/tests/test_security_matrix.py -q` -- all 40 pass, no regressions
- [ ] CI

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GitHub API retry handling, including support for server errors and capped retry delays.
  * Prevented unexpected API errors from being incorrectly reported as compliant branch protection.
  * Marked security checks as errors when repository results are incomplete due to access restrictions or rate limits.
  * Correctly evaluates unprotected default branches as allowing force pushes while excluding unavailable, rate-limited results.

### DIFF
--- a/packages/checks/src/checks/github_checks.py
+++ b/packages/checks/src/checks/github_checks.py
@@ -25,10 +25,30 @@ def _is_secondary_rate_limit(r: httpx.Response) -> bool:
         return False
     return "Retry-After" in r.headers or r.headers.get("X-RateLimit-Remaining") == "0"
 
+# Cap on honoring a server-supplied Retry-After value, so a single malformed/huge value
+# can't block a scan for an unreasonable amount of time -- GitHub's real rate-limit
+# Retry-After values are ordinarily well under this.
+_MAX_RETRY_AFTER_SECONDS = 60
+
+
+def _retry_delay_seconds(r: httpx.Response, attempt: int) -> float:
+    """Prefer the server's own Retry-After value (GitHub's rate-limit responses include
+    one) over a blind exponential backoff -- retrying before the window it asked for just
+    burns the fixed attempt budget hitting the same limit again."""
+    raw = r.headers.get("Retry-After")
+    if raw is not None:
+        try:
+            return min(float(raw), _MAX_RETRY_AFTER_SECONDS)
+        except ValueError:
+            pass
+    return 2**attempt
+
 
 def _get_with_retry(client: httpx.Client, url: str, headers: dict) -> httpx.Response:
     # Same retry/backoff contract as GitHubClient.request (apps/api/src/services/github_client.py):
-    # 3 attempts, exponential backoff on connection errors, a 429, or a rate-limit-flavored 403.
+    # 3 attempts, exponential backoff on connection errors, a 429, or a rate-limit-flavored 403 --
+    # plus a 5xx, which is presumed transient (GitHub-side issue) same as everywhere else in this
+    # codebase that talks to the GitHub API.
     for attempt in range(3):
         try:
             r = client.get(url, headers=headers)
@@ -37,8 +57,8 @@ def _get_with_retry(client: httpx.Client, url: str, headers: dict) -> httpx.Resp
                 time.sleep(2**attempt)
                 continue
             raise
-        if (r.status_code == 429 or _is_secondary_rate_limit(r)) and attempt < 2:
-            time.sleep(2**attempt)
+        if (r.status_code == 429 or _is_secondary_rate_limit(r) or r.status_code >= 500) and attempt < 2:
+            time.sleep(_retry_delay_seconds(r, attempt))
             continue
         return r
     raise RuntimeError("request loop exhausted without returning")
@@ -91,9 +111,12 @@ def _branch_protection_status(exc: httpx.HTTPStatusError) -> str:
     code = exc.response.status_code
     if code == 404:
         return "unprotected"
-    if code in (403, 429):
-        return "unknown"
-    return "unprotected"
+    # 403/429 (no access / rate-limited) and any other unexpected status (a transient
+    # 5xx, most notably) all mean branch-protection status genuinely can't be evaluated
+    # right now -- falling through to "unprotected" here previously flipped an org's
+    # entire branch-protection check to "fail" on a flaky GitHub 5xx even when every
+    # real repo was protected.
+    return "unknown"
 
 
 class OrgMFARequired(Check):
@@ -238,6 +261,13 @@ class DependabotAlertsCheck(Check):
         if forbidden == len(repos):
             return {"status": "error", "value": counts}
         compliant = counts["critical"] == 0 and counts["high"] == 0
+        if compliant and forbidden > 0:
+            # 403 means unknown, not zero -- a clean result from only the *reachable*
+            # repos must not be reported as "pass" while some repos' real alert counts
+            # are still unknown. A "fail" from repos we could actually see stays valid
+            # regardless (a real problem was found), so only the false-clean case needs
+            # downgrading here.
+            return {"status": "error", "value": counts}
         return {"status": "pass" if compliant else "fail", "value": counts}
 
 
@@ -285,7 +315,13 @@ class CodeScanningCheck(Check):
         value = {"open": open_count, "repos_with_alerts": repos_with_alerts, "total_repos": total_repos}
         if forbidden == total_repos:
             return {"status": "error", "value": value}
-        return {"status": "pass" if open_count == 0 else "fail", "value": value}
+        compliant = open_count == 0
+        if compliant and forbidden > 0:
+            # Same reasoning as DependabotAlertsCheck: a clean result from only the
+            # reachable repos must not be reported as "pass" while some repos' real
+            # alert status is still unknown (403 means unknown, not zero).
+            return {"status": "error", "value": value}
+        return {"status": "pass" if compliant else "fail", "value": value}
 
 
 class DefaultBranchNoForcePushCheck(Check):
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
@coderabbitai review this pr

nazarli-shabnam (commented):
Confirmed — all 5 misclassification bugs from https://github.com/nazarli-shabnam/clevis/issues/245 are fixed point-for-point (404 handling, unknown-vs-unprotected fallback, forbidden-repo escalation for Dependabot/CodeScanning, 5xx retry, real Retry-After value), with regression tests that replace rather than just extend the old buggy assertions. CI green.

PR #264, opened 2026-07-24
fix(api): resolve an org's GitHub App installation from personal endpoints

nazarli-shabnam:
## Summary
Overview's cockpit/my-view widgets, plus the personal cache/security/automation endpoints, called `resolve_personal_token` exclusively — which only ever matches an installation scoped to the caller's own user id. An org-only GitHub App installation (`owner_user_id=NULL`, `org_id` set) could never be found there, so a correctly connected org still showed "No GitHub App installation found" everywhere except the org-scoped Activity/analytics endpoints.

- Adds `resolve_owner_token` (`token_resolution.py`): if the requested `owner` is a Clevis org the caller is at least a member of, prefer that org's installation (same resolution the `/orgs/{org_login}/...` endpoints use); otherwise fall back to the existing personal-token path unchanged.
- Membership-gated so a random authenticated user can't pull another org's GitHub data through a `/me/*` endpoint just by typing its login.
- Role-gated (`min_role="admin"`) for the two privileged actions — cache clear and workflow dispatch — matching what their org-scoped equivalents already require.

**Touches RBAC-adjacent token-resolution code** across `analytics.py`, `actions_cache.py`, `automation.py`, and `security.py`. No schema/migration change.

Also included:
- The github-callback page now shows which account/org actually got connected in its success message, instead of a generic "Connected" — this is what surfaced the org-installation bug in the first place (a personal-account install succeeded silently while the intended org went unconnected).
- A small `self-hosting.md` clarification: the App's private key is a separate secret from the webhook secret and isn't on the App's creation form — it's generated afterward from the App's own settings page.

(Note: this PR previously contained an unrelated third-party Claude Code skill addition, split out to #265.)

## Test plan
- [x] `pytest -q --cov=... --cov-fail-under=85` — 565 passed
- [x] `diff-cover` against `origin/main` — 100% on all changed lines
- [x] `bun run check` (typecheck + lint + build) on the UI change
- [x] Added tests for `resolve_owner_token`: org-member success, non-connected-org fallback, non-member rejection, admin-gate rejection/success

## Summary by CodeRabbit

* **New Features**
  * GitHub connection confirmation now displays the connected account login and whether it’s a personal or organization account.
  * Personal GitHub features can now use organization access when you’re a member with sufficient permissions.

* **Bug Fixes**
  * Improved token selection for analytics, automation, security, and cache actions.
  * Privileged personal actions now correctly return **403** when your org role is insufficient.

* **Documentation**
  * Self-hosting guide clarified GitHub App permissions and the correct setup of the private key secret vs the webhook secret.

* **Tests**
  * Added regression coverage for member/admin permission enforcement on personal endpoints.


nazarli-shabnam:
Went through the CodeRabbit findings. 5 of the 8 referenced karpathy-skill files (README.md, examples/, references/research/) that had already moved to #265 during the branch restructure — stale/inapplicable here, not fixed on this PR.

3 were valid and fixed (`01fba5a`):
- **Case-insensitive org lookup**: `resolve_owner_token` used `org_repo.get_by_login` (exact match) while `installation_repo.get_for_org` is already case-insensitive (#246) — a casing variant of a connected org's login would silently fall through to personal-token resolution. Added `get_by_login_ci`.
- **Real security gap**: a member (not admin) of a connected org could bypass the admin-only gate on cache-clear/workflow-dispatch by supplying their own PAT — `resolve_owner_token` fell through to `resolve_personal_token` on an insufficient role, which honors a client-supplied token. Now raises `InsufficientOrgRole` → 403 instead of falling through, whenever a membership row exists but is under-ranked.
- **Redirect timer bug** on the callback page: `useEffect` depended on `router`, whose object identity isn't stable across renders — every state update in `run()` re-triggered the cleanup, cancelling the just-scheduled redirect before it fired. Dropped `router` from the deps (matches the existing eslint-disable convention already used for this in `cache-panel.tsx`).

All 3 have regression tests, 569 tests pass, 100% diff coverage, CI green.

nazarli-shabnam:
@coderabbitai review again this PR

PR #295, opened 2026-07-25
fix(auth): reject passwords over 72 bytes instead of silently truncating

### LINKED ISSUE #269: fix(auth): passwords over 72 bytes crash register/setup/login with an unhandled 500
## Problem

`_hash_password`/`_verify_password` in `apps/api/src/routers/auth.py` call `bcrypt.hashpw`/`bcrypt.checkpw` directly on the raw password bytes, with only a *minimum* length check (`_MIN_PASSWORD_LEN = 12`) — there's no upper bound. `bcrypt` (pinned in `apps/api/requirements.txt`) raises `ValueError: password cannot be longer than 72 bytes` for any input over 72 bytes, and none of `/auth/setup`, `/auth/register`, or `/auth/login` catch it.

## Failure scenario

Any unauthenticated client submits a password longer than 72 bytes to `/auth/register`, `/auth/setup`, or `/auth/login`. The `ValueError` propagates unhandled, producing a 500 instead of a normal 4xx validation error. This is trivial to trigger with no auth required, and depending on deployment error-handling config could leak stack-trace/framework detail in the response.

It also affects `login()`'s timing-safety branch: when no matching user is found, the code is expected to still run `bcrypt.checkpw` against a dummy hash so response timing doesn't reveal whether the email exists. If the submitted password is >72 bytes, that dummy-hash call itself throws instead of completing, breaking the timing-safety property in exactly the input range where it's least expected.

## Why this matters

- Unauthenticated 500s from oversized input are a minor DoS-friendly footgun and noisy in error tracking.
- The timing-safety invariant silently breaks for the >72-byte case.
- Related: bcrypt truncates at 72 bytes internally with no user-facing error today (once the ValueError is fixed, this should be replaced with an explicit "password too long" validation error before hashing — not silent truncation — so two different long passwords sharing the same first 72 bytes aren't treated as identical).

## Severity

Medium — easy to trigger, unauthenticated, on multiple hot paths; low direct exploitability but a real correctness/DoS gap that should be closed before wider traffic hits these endpoints.

nazarli-shabnam:
## Summary

Fixes #269. `_hash_password`/`_verify_password` in `apps/api/src/routers/auth.py` call `bcrypt.hashpw`/`bcrypt.checkpw` directly on raw password bytes, with only a lower-bound length check (`_MIN_PASSWORD_LEN = 12`) and no upper bound.

**Correction from the issue's original framing** (caught during code review, verified directly against the pinned dependency): `bcrypt==4.2.1` — the version actually pinned in `apps/api/requirements.txt` — does **not** raise `ValueError` for passwords over 72 bytes. That behavior was only added in `bcrypt>=5.0`. Verified with `bcrypt.checkpw()` directly: two different passwords sharing the same first 72 bytes hash and verify identically under 4.2.1 today. So the real bug on this pinned version is **silent password-collision via truncation**, not an unhandled 500 — the issue itself flagged this as a secondary concern ("bcrypt truncates at 72 bytes internally with no user-facing error today... should be replaced with an explicit 'password too long' validation error"), and that's the actual live bug here.

The fix itself doesn't change: reject oversized passwords with a clean validation error before hashing, rather than letting bcrypt silently truncate them.

## Why this approach

- **`setup()`/`register()`**: added an explicit `_MAX_PASSWORD_LEN = 72`-byte check (mirroring the existing `_MIN_PASSWORD_LEN` 422 pattern) *before* hashing, so oversized passwords get a clean validation error and can never produce a hash derived from silently-truncated input.
- **`login()`**: can't reject with 422 (unauthenticated clients shouldn't get a different status code that reveals anything), and must preserve the existing timing-safety property (`_verify_password` always runs a real bcrypt call so response timing doesn't reveal whether an email exists). Since `setup()`/`register()` now guarantee no real stored password can exceed 72 bytes, an oversized login attempt is always a wrong guess — `_verify_password` truncates the input to 72 bytes for a bounded bcrypt call (still pays the real cost against the actual stored hash) and unconditionally returns `False`, rather than relying on bcrypt's implicit truncation behavior to (coincidentally) reject it.

## Test plan
- [x] Added regression tests: `test_setup_rejects_password_over_72_bytes`, `test_register_rejects_password_over_72_bytes`, `test_login_rejects_password_over_72_bytes`
- [x] Full backend suite: `pytest -q` — 599 passed
- [x] Manually verified against the pinned `bcrypt==4.2.1`: `bcrypt.checkpw(b"a"*72+b"AAAA", hash_of(b"a"*72+b"BBBB"))` returns `True` on unpatched behavior, confirming the real collision bug
- [x] `apps/ui` unaffected (backend-only change); pre-commit `tsc --noEmit` passed

No migration, no other sensitive-file changes beyond the auth router itself.

### DIFF
--- a/apps/api/src/routers/auth.py
+++ b/apps/api/src/routers/auth.py
@@ -47,6 +47,9 @@ logger = logging.getLogger(__name__)
 router = APIRouter()
 
 _MIN_PASSWORD_LEN = 12
+# bcrypt's hard limit -- bcrypt.hashpw/checkpw raise ValueError on any input over 72 bytes
+# rather than truncating, so this must be validated before either is ever called.
+_MAX_PASSWORD_LEN = 72
 _VERIFY_TOKEN_TTL = timedelta(hours=24)
 
 
@@ -83,10 +86,19 @@ def _hash_password(password: str) -> str:
 
 
 def _verify_password(password: str, password_hash: str | None) -> bool:
-    if not password_hash:
-        bcrypt.checkpw(password.encode(), _DUMMY_PASSWORD_HASH.encode())
+    password_bytes = password.encode()
+    # No account's real password can be over _MAX_PASSWORD_LEN bytes -- setup()/register()
+    # reject it before hashing -- so an oversized guess is always wrong. Still run a real
+    # (truncated) bcrypt call rather than short-circuiting, so response timing doesn't
+    # reveal the oversized-password case any more than it reveals a nonexistent email below.
+    if len(password_bytes) > _MAX_PASSWORD_LEN:
+        password_bytes = password_bytes[:_MAX_PASSWORD_LEN]
+        bcrypt.checkpw(password_bytes, (password_hash or _DUMMY_PASSWORD_HASH).encode())
         return False
-    return bcrypt.checkpw(password.encode(), password_hash.encode())
+    if not password_hash:
+        bcrypt.checkpw(password_bytes, _DUMMY_PASSWORD_HASH.encode())
+        return False
+    return bcrypt.checkpw(password_bytes, password_hash.encode())
 
 
 # ── Schemas ───────────────────────────────────────────────────────────────────
@@ -201,6 +213,11 @@ def setup(body: SetupRequest, db: Session = Depends(get_db)):
             status_code=422,
             detail=f"Password must be at least {_MIN_PASSWORD_LEN} characters",
         )
+    if len(body.password.encode()) > _MAX_PASSWORD_LEN:
+        raise HTTPException(
+            status_code=422,
+            detail=f"Password must be at most {_MAX_PASSWORD_LEN} bytes",
+        )
     user = User(
         email=body.email,
         name=body.name,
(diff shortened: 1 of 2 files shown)


nazarli-shabnam:
### Code review

Found 1 issue (already fixed in this PR):

1. The original commit's comments, test docstrings, PR description, and commit message all asserted that `bcrypt.hashpw`/`bcrypt.checkpw` raise `ValueError` for passwords over 72 bytes. Verified directly against the actual pinned dependency (`bcrypt==4.2.1`) — this is false. That version silently truncates rather than raising; `ValueError` was only added in `bcrypt>=5.0`. Confirmed the real consequence instead: `bcrypt.checkpw(b"a"*72+b"BBBB", hash_of(b"a"*72+b"AAAA"))` returns `True` under 4.2.1 — two different passwords sharing the same first 72 bytes hash and verify identically. This is an unverified factual claim baked into code comments and tests (AGENTS.md: "Verify before asserting... don't infer from naming conventions or from how a 'similar' codebase might work").

This has been corrected in [c4320dc](https://github.com/nazarli-shabnam/clevis/blob/c4320dcf4b547921e26391efb895d30336afb32b/apps/api/src/routers/auth.py#L49-L56): the comments/tests/PR description now describe the actual bug (silent truncation causing password collisions), not a crash. The fix itself (reject >72-byte passwords with 422 before hashing) was already correct — only the stated justification was wrong.

PR #296, opened 2026-07-25
fix(auth): apply rate limiting to /auth/setup

### LINKED ISSUE #279: /auth/setup has no rate limit unlike sibling auth endpoints
`apps/api/src/routers/auth.py:185` — `/auth/setup` has no `dependencies=[Depends(rate_limit())]`, while `register`, `login`, `verify-email`, and `resend-verification` all do.

It's guarded by a Postgres advisory lock and becomes a no-op 409 after the first admin is created, so real-world severity is likely low. Flagging because it's an unauthenticated POST that runs password-validation logic on every call before the lock/count check, on every request until first-run setup completes.

Low confidence this needs any change — worth a second look, may be closed as won't-fix if the advisory lock is judged sufficient protection.

nazarli-shabnam:
## Summary

Fixes #279. `/auth/setup` (`apps/api/src/routers/auth.py:185`) was missing `dependencies=[Depends(rate_limit())]`, unlike every sibling auth endpoint (`register`, `login`, `verify-email`, `resend-verification`).

## Why this matters

`/auth/setup` is guarded by a Postgres advisory lock (`pg_advisory_xact_lock`) plus a `count() > 0` check, but that only protects against two *concurrent* setup calls racing to create the first admin (a correctness/race safeguard) -- it does nothing to throttle a single caller hitting the unauthenticated endpoint repeatedly before first-run setup completes. Since setup becomes a no-op 409 after the first admin exists, real-world severity is low, but it's an unauthenticated POST that runs password-validation logic on every call, with no rate limit unlike its siblings.

## Change

One-line addition: `dependencies=[Depends(rate_limit())]` on the `/setup` route decorator, matching the existing pattern used by `register`/`login`/`verify-email`/`resend-verification`. No change to the advisory-lock logic itself -- the two mechanisms address different problems (race safety vs. abuse throttling) and both are needed.

## Test plan
- [x] Added `test_setup_applies_a_per_ip_rate_limit`, mirroring the existing `test_rate_limit.py` pattern: exceeds the default `max_requests=10` and asserts a 429 on the 11th call.
- [x] Full backend suite: `pytest -q` -- 599 passed.

No migration, no other sensitive-file changes beyond the auth router itself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added per-IP rate limiting to the authentication setup endpoint.
  * Excessive setup requests now return a rate-limit response, helping protect the endpoint from abuse.

* **Tests**
  * Added coverage to verify rate limits are enforced consistently, including after unsuccessful setup attempts.

### DIFF
--- a/apps/api/src/routers/auth.py
+++ b/apps/api/src/routers/auth.py
@@ -182,7 +182,7 @@ def setup_required(db: Session = Depends(get_db)):
     return {"setup_required": count == 0}
 
 
-@router.post("/setup", response_model=TokenResponse, status_code=status.HTTP_201_CREATED)
+@router.post("/setup", response_model=TokenResponse, status_code=status.HTTP_201_CREATED, dependencies=[Depends(rate_limit())])
 def setup(body: SetupRequest, db: Session = Depends(get_db)):
     """First-run only. Returns 409 if any user already exists."""
     # Two concurrent /auth/setup calls with *different* emails can both pass the
--- a/apps/api/tests/test_auth.py
+++ b/apps/api/tests/test_auth.py
@@ -91,6 +91,20 @@ def test_setup_rejects_duplicate(auth_client):
     assert resp.status_code == 409
 
 
+def test_setup_applies_a_per_ip_rate_limit(auth_client):
+    # Regression test for issue #279: /auth/setup was missing the rate_limit() dependency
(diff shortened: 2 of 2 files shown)


nazarli-shabnam:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #297, opened 2026-07-25
fix(auth): enforce case-insensitive email uniqueness

### LINKED ISSUE #268: fix(auth): case-sensitive email uniqueness lets one mailbox back multiple accounts
## Problem

`User.email` (`apps/api/src/core/db.py`) is a case-sensitive column, and every place that checks "does an account with this email already exist" compares it case-sensitively: `register()` and `login()` in `apps/api/src/routers/auth.py`, and `find_or_create_user()`'s `EmailAlreadyRegistered` check in `apps/api/src/routers/github_auth.py`. Real mailboxes are effectively case-insensitive (almost all providers ignore case in the domain, and typically the local-part too), so this DB-level invariant doesn't match reality.

## Failure scenario

A user registers `alice@example.com`. Later, the same person (or an attacker who doesn't control the inbox but wants to test the boundary) registers `Alice@example.com`. The API creates a second, fully independent account — `User.email == body.email` doesn't match the existing row, and there's no case-insensitive unique constraint to reject it at the DB layer either.

This directly undermines the invariant the email-verification work for #217 relies on ("an account with this email" is unique, so verifying the inbox is enough to trust invite-acceptance). With two accounts for one mailbox:
- `accept_invitation`'s email-match check can pass for the "wrong" (differently-cased) account if that's the one currently signed in, since it also compares case-sensitively.
- `login()`'s rate-limit key (`body.email.lower()`) doesn't line up with which DB row actual password guesses are hitting, weakening the rate limit's effectiveness per-account.
- Pending-invitation lookups (`_pending_invitations_for`) can silently miss or duplicate depending on which-case account owns the invite.

## Why this matters

This is the same class of identity-confusion bug the #217 email-verification work was meant to close, but through a code path that work didn't cover — case-insensitive email handling. Fixing it likely means: normalizing email to lowercase (or a citext/functional-index approach) at write time everywhere an account is created or matched by email, backed by a real case-insensitive unique constraint rather than only an application-level check-then-insert.

## Severity

Medium — requires either attacker control of a mailbox provider's case-folding behavior or a confused/duplicate legitimate user, but breaks a security invariant other code explicitly relies on.

nazarli-shabnam:
## Summary

Fixes #268. `User.email` (`apps/api/src/core/db.py`) had a case-sensitive unique column, and every place that checks "does an account with this email already exist" compared case-sensitively too: `register()` and `login()` in `apps/api/src/routers/auth.py`, and `find_or_create_user()`'s `EmailAlreadyRegistered` check in `apps/api/src/routers/github_auth.py`. Real mailboxes are effectively case-insensitive, so `alice@example.com` and `Alice@example.com` could register as two fully independent accounts.

This undermines the identity-uniqueness invariant the #217 email-verification work relies on ("an account with this email" is unique), and diverges from `invitations`, which already compares emails case-insensitively (`ilike`, `.lower()`) via `invitation_repo.py`.

## Schema change (please review carefully)

**This PR includes a new migration: `apps/api/alembic/versions/0019_case_insensitive_email.py`.**

- Lowercases all existing `users.email` values.
- Drops the plain `users_email_key` unique constraint and replaces it with a unique index on `lower(email)`, since Postgres can't express case-insensitive uniqueness through a plain column constraint.
- **Data-loss/backfill risk**: if two *different* existing accounts already share an email address differing only by case, lowercasing would collide. The migration detects this up front (`GROUP BY lower(email) HAVING COUNT(*) > 1`) and **raises before making any change** rather than silently merging or dropping either account — verified locally by manually inserting a colliding pair under the old constraint and confirming `alembic upgrade head` aborts cleanly with a clear error naming the offending email, leaving the DB untouched. No such collision is expected in practice (case-sensitive uniqueness has been enforced since migration 0004), but this guards against the case where it wasn't, or data was loaded out of band.
- Added a matching `Index("uq_users_email_lower", text("lower(email)"), unique=True)` to the `User` model's `__table_args__` so `alembic check` sees no drift (verified: `alembic check` → "No new upgrade operations detected").
- Verified both `alembic upgrade head` and `alembic downgrade -1` apply cleanly.

## Code changes

- `auth.py` `register()`/`login()` and `github_auth.py` `find_or_create_user()` now compare via `func.lower(User.email) == email.lower()` instead of exact match.
- New accounts are stored with `email.lower()` at insert time (`setup()`, `register()`, `find_or_create_user()`), so stored values read consistently going forward.

## Test plan
- [x] New regression tests: `test_register_rejects_duplicate_email_different_case`, `test_register_stores_email_lowercased`, `test_setup_stores_email_lowercased`, `test_login_valid_with_different_case_email`, `test_refuses_to_auto_link_when_existing_email_differs_only_by_case`
- [x] Full backend suite: `pytest -q` — 601 passed
- [x] Manually verified the migration's collision guard (see above) and both migration directions locally

Touches sensitive auth code (email matching/uniqueness) and includes a schema migration — flagging both per the repo's guardrails.

## Summary by CodeRabbit

* **New Features**
  * Email addresses are now handled case-insensitively across account setup, registration, login, and GitHub sign-in.
  * New accounts store email addresses in lowercase for consistent account management.

* **Bug Fixes**
  * Prevented duplicate accounts when emails differ only by capitalization.
  * Login now succeeds regardless of email capitalization.
  * GitHub sign-in no longer incorrectly links or creates accounts for case-only email variations.

* **Tests**
  * Added regression coverage for case-insensitive email behavior across auth flows.


nazarli-shabnam:
### Code review

No issues found. Checked migration safety (collision-detection guard, upgrade/downgrade correctness, ORM/migration drift), all three email-comparison call sites for consistency, interaction with prior email-verification/invitation work (#217, #218, #248), and CLAUDE.md/AGENTS.md compliance (migration guardrail, sensitive-file flagging, scope).

nazarli-shabnam:
### Follow-up: fixed 1 CodeRabbit finding

The migration's collision-guard exception embedded every conflicting email address verbatim, which could leak PII into deployment/Alembic logs. Now reports only the collision count, with a query an operator can run directly against the DB to find the actual addresses.

Verified locally: downgrade/upgrade cycle still applies cleanly, and manually inserting a case-variant collision under the pre-0019 constraint still correctly raises before making any change — confirmed the new message ("1 email address(es) are shared by multiple existing users...") with no raw emails included.

PR #299, opened 2026-07-25
feat(ui): hide GitHub token field when App installation covers owner

### LINKED ISSUE #276: Hide "optional GitHub token" field when a GitHub App is already installed
The hint "optional if the GitHub App is connected for this org" (or equivalent) is shown alongside a token input on 4 pages, regardless of whether an installation actually covers the org/account in view:

- `apps/ui/app/automation/page.tsx:142`
- `apps/ui/app/repos/page.tsx:266`
- `apps/ui/app/security/page.tsx:229`
- `apps/ui/components/repo/cache-panel.tsx:130`

None of these currently check installation state. `api.installations.list()` already exists and is used for this purpose elsewhere (`apps/ui/components/app-sidebar.tsx:213-215`, `apps/ui/app/settings/page.tsx:237`).

**Decision (confirmed with product owner):** when an installation covers the currently active scope, hide the token field entirely on these pages rather than just leaving it visible with a hint. Reuse the existing `["installations"]` query so this doesn't introduce a duplicate fetch.

nazarli-shabnam:
## Summary

Fixes #276. The hint "optional if the GitHub App is connected for this org" was shown alongside a token input on 4 pages regardless of whether an installation actually covers the org/account in view:

- `apps/ui/app/automation/page.tsx`
- `apps/ui/app/repos/page.tsx`
- `apps/ui/app/security/page.tsx`
- `apps/ui/components/repo/cache-panel.tsx`

None of these checked installation state. `api.installations.list()` already exists and is used for this purpose elsewhere (`apps/ui/components/app-sidebar.tsx`, `apps/ui/app/settings/page.tsx`).

## Change

Per the confirmed product decision in the issue: when an installation covers the currently active scope, hide the token field entirely rather than just leaving it visible with a hint.

- Wired the existing `["installations"]` query into all 4 sites (React Query dedupes by key, so this doesn't introduce a duplicate fetch — it reuses whatever `app-sidebar.tsx`/`settings/page.tsx` already fetched).
- Each page compares its own resolved `owner` state (not necessarily the sidebar's active scope — the user can overtype it on automation/repos/security) against `installs.some(i => i.account_login === owner)`.
- `cache-panel.tsx` gets `owner`/`repo` as props rather than from `useActiveScope`, so it wires the same query directly.

## Test plan
- [x] `bun run check` (typecheck + lint + build) — passes
- [x] Added `installations.list` mocks to all 4 affected test files (`automation-page.test.tsx`, `repos-page.test.tsx`, `security-page.test.tsx`, `repo-detail-page.test.tsx`) — **none of them previously mocked this endpoint**, so `api.installations.list()` was silently failing/falling back to an empty array in every existing test run, masking the untested code path entirely.
- [x] Added a regression test per page verifying the token field is hidden when an installation covers the entered owner, and (for automation) that it's still shown when none does.
- [x] `bun run test` — all 4 affected test files pass (72 tests)

No backend changes.

## Summary by CodeRabbit

- **New Features**
  - “GitHub Token” inputs are now hidden when a GitHub App installation exists for the chosen owner, using both personal and org-scoped installation lookups.
  - Updated automation, repos, security, and actions cache views to reflect installation status.

- **Bug Fixes**
  - Token auto-resolution no longer applies resolved legacy/saved tokens when an installation already exists for the current owner.

- **Tests**
  - Added/extended UI tests to verify token visibility across installation present/absent and org-lookup error scenarios.

- **Chores**
  - Added an org-scoped installations API method (`listForOrg`).

### DIFF
--- a/apps/ui/app/automation/page.tsx
+++ b/apps/ui/app/automation/page.tsx
@@ -48,12 +48,28 @@ export default function AutomationPage() {
     queryKey: ["installations"],
     queryFn: () => api.installations.list(),
   })
-  const hasInstallationForOwner = installs.some((i) => i.account_login === owner)
+  // list() above only covers the caller's *personal* installations -- an org's App
+  // installation requires the separate org-scoped endpoint. Errors (403/404, e.g. the
+  // org isn't a recognized Clevis org yet) are treated as "not installed" rather than
+  // surfaced, matching this query's only purpose here (a soft signal to hide the token
+  // field, not something the user needs an error for).
+  const orgInstallsQuery = useQuery<InstallationMeta[]>({
+    queryKey: ["installations.org", owner.trim()],
+    queryFn: () => api.installations.listForOrg(owner.trim()),
+    enabled: owner.trim().length > 0,
+    retry: false,
+  })
+  const hasInstallationForOwner =
+    installs.some((i) => i.account_login === owner) || (orgInstallsQuery.data?.length ?? 0) > 0
 
   const resolveMutation = useMutation({
     mutationFn: (org: string) => api.tokens.resolve(org),
     onSuccess: (data, org) => {
-      if (shouldApplyResolvedToken(org, owner)) {
+      // Skip applying a legacy saved token once an installation covers this owner --
+      // otherwise it'd be silently used (the token field, and its "saved" indicator,
+      // are hidden in that case) and could override the installation-token path the
+      // hidden field implies is now authoritative.
+      if (shouldApplyResolvedToken(org, owner) && !hasInstallationForOwner) {
         setToken(data.token)
         setTokenSaved(true)
       }
--- a/apps/ui/app/automation/page.tsx
+++ b/apps/ui/app/automation/page.tsx
@@ -1,7 +1,7 @@
 "use client"
 
 import { useEffect, useState } from "react"
-import { useMutation } from "@tanstack/react-query"
+import { useMutation, useQuery } from "@tanstack/react-query"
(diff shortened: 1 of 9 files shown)


nazarli-shabnam:
### Follow-up: fixed a critical gap found in my own review

The original implementation only queried `GET /me/installations` (`api.installations.list()`), which returns exclusively the caller's *personal* installations — org-connected App installations live under the separate `GET /orgs/{org}/installations` endpoint. That meant `hasInstallationForOwner` was effectively always `false` for org owners, so the token field never actually hid for the primary use case this issue describes (orgs with a connected GitHub App).

Fixed in 207f7c9:
- Added `api.installations.listForOrg(orgLogin)` in [`apps/ui/lib/api/client.ts`](https://github.com/nazarli-shabnam/clevis/blob/207f7c9336f2f6877964d92e9faca5c6c1ce8118/apps/ui/lib/api/client.ts)
- Queried it alongside the existing personal-installations query on all four pages (`automation`, `repos`, `security`, `cache-panel`), OR'd together for `hasInstallationForOwner`. A 403/404 (not an org member / org not connected) is treated as "not installed" via `retry:false`, falling back to showing the token field.
- Guarded `resolveMutation`'s `onSuccess` so a resolved legacy saved token is never silently applied once an installation is detected — it previously could set `tokenSaved=true` behind the now-hidden field.

Added regression tests for both the org-level-hide case and the error-fallback case to all four test files. Full suite (366 tests), typecheck, and lint all pass locally.

Issue #86, opened 2026-07-10
fix(api): failed-job error strings stored and exposed verbatim with no redaction


nazarli-shabnam:
## Problem

`mark_failed()` (`apps/api/src/repositories/job_repo.py:39-43`) stores the raw exception string in `Job.result` on failure, which is later returned as-is by `GET /jobs`. If the worker's GitHub API call fails with an error message that echoes request details (URL, query params, headers), that text is persisted and exposed via the API to any workspace admin with no sanitization or redaction.

## Failure scenario

A GitHub API error response includes some detail from the outbound request (e.g. a malformed param value) in its error text; that raw string is stored and later rendered in the `/jobs` admin view verbatim, as an unbounded/unsanitized surface flowing from an external HTTP response into stored, API-exposed data.

## Suggested fix

Truncate and/or sanitize exception strings before storing them in `Job.result` — cap length and strip anything that looks like a token/credential fragment as defense in depth.

## Severity

Low — endpoint is already workspace-admin-gated, but the storage/exposure pattern is unbounded and unsanitized.

PR #301, opened 2026-07-25
fix(ui,worker): support per-cache-key clear and surface GitHub's error body

### LINKED ISSUE #274: Actions cache clearing silently fails — missing key param, plus buried/confusing UX
Root cause confirmed by tracing the full flow: `apps/ui/components/repo/cache-panel.tsx` never collects a cache `key` (or `ref`), even though the cache table already lists individual `key` values per row.

Flow: UI → `POST /orgs/{org}/repos/{owner}/{repo}/actions-caches/clear` (`apps/api/src/routers/actions_cache.py:59`) → `cache_service.clear()` → enqueues job `github.clear_actions_cache` → worker `_handle_clear_actions_cache` (`apps/worker/src/worker.py:155-202`), which calls `DELETE /repos/{owner}/{repo}/actions/caches` with an **empty** `params` object.

GitHub's actual delete-cache endpoint requires `key` as a query parameter — there's no bulk-delete-all-caches API. So every clear job issued today is expected to fail, which matches the reported symptom: Activity shows a failed job right after clearing.

Secondary issue: on failure, `jobs.result` only stores the HTTP status code (`worker.py:198`), not GitHub's actual error body — making failures hard to diagnose from the Activity page.

There's also a UX/discoverability complaint: this action is currently buried in a repo-detail tab, which is a nonobvious place to find it.

**Ask:**
- Let the user pick a specific cache row to clear (using its `key`), and actually pass `key` through the full chain: UI → API payload → job params → worker request.
- Surface GitHub's real error message in `jobs.result` instead of just the status code.
- Revisit where this action lives in the nav/IA given it's hard to find today.

nazarli-shabnam:
## Summary

Fixes #274.

- **UI**: The Actions Cache panel's "Clear" button always cleared *every* cache for a repo, even though the backend (`CacheClearInput.key`/`.ref`) and the table itself (each row already shows its `key`) fully supported scoping to one entry. Added a per-row "Clear key" action that scopes the same `clearMutation` to that row's `{ key, ref }`, using the same two-click "arm then confirm" pattern as the existing global Clear button. The global button is unchanged and still clears everything when no row is targeted.
- **Worker**: `_mark_failed`/`_requeue_for_retry` only ever recorded the bare HTTP status code in `jobs.result` (e.g. `"GitHub API error: 403"`). Added a `_github_error_message` helper that reads GitHub's JSON error body's `message` field when present, so a failed job's audit trail shows *why* it failed (e.g. `"Resource not accessible by integration"`) instead of just a status code.
- **Nav**: confirmed the standalone `/repos/{repo}/cache` route is already linked from the repos list page's per-row "Cache" link (`apps/ui/app/repos/page.tsx:161`) — not orphaned, no change needed there.

No backend/schema change was required — `CacheClearInput`, `cache_service.clear()`, and the worker's `_handle_clear_actions_cache` already plumbed `key`/`ref` through end-to-end; only the UI never collected them.

## Why

Without a way to clear a single stale cache entry, the only option in the UI was "clear everything for this repo," which is unnecessarily destructive for the common case of wanting to evict one bad/stale cache key. And a failed clear job's error message ("GitHub API error: 403") gave an operator no way to tell *why* it failed without going to GitHub's own UI.

## Test plan
- [x] `bun run typecheck`, `bun run lint`, full `bunx vitest run` (354 tests) all pass locally
- [x] New test: "clears a single row's cache via its own key-scoped Clear key action, without affecting others" in `cache-page.test.tsx`, asserting the row action calls `api.cache.clear` with `key`/`ref` set and doesn't arm the global button
- [x] `apps/worker/tests/test_process_job.py` unit tests (mocked httpx) pass — 13/15 pass; 2 unrelated reclaim/DB-integration tests error locally only because no local Postgres is running (pre-existing environment limitation, not caused by this change)

### DIFF
--- a/apps/ui/components/repo/cache-panel.tsx
+++ b/apps/ui/components/repo/cache-panel.tsx
@@ -32,6 +32,9 @@ export function CachePanel({ owner, repo, active = true }: CachePanelProps) {
   const [tokenSaved, setTokenSaved] = useState(false)
   const [actor, setActor] = useState("")
   const [clearArmed, setClearArmed] = useState(false)
+  // null = clearing every cache for this repo (the existing global buttons); set to a
+  // specific { key, ref } when the user clicks a row's own "Clear" action instead.
+  const [clearTarget, setClearTarget] = useState<{ key: string; ref: string } | null>(null)
 
   // Auto-resolve saved token for this owner
   const resolveMutation = useMutation({
@@ -74,6 +77,7 @@ export function CachePanel({ owner, repo, active = true }: CachePanelProps) {
     listMutation.reset()
     clearMutation.reset()
     setClearArmed(false)
+    setClearTarget(null)
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [owner, repo])
 
@@ -97,7 +101,13 @@ export function CachePanel({ owner, repo, active = true }: CachePanelProps) {
 
   const clearMutation = useMutation({
     mutationFn: (dryRun: boolean) =>
-      api.cache.clear(owner, repo, { token, actor, dry_run: dryRun }),
+      api.cache.clear(owner, repo, {
+        token,
+        actor,
+        dry_run: dryRun,
+        key: clearTarget?.key,
+        ref: clearTarget?.ref,
+      }),
   })
 
   const isLoading = listMutation.isPending || clearMutation.isPending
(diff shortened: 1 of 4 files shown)


nazarli-shabnam:
### Code review

Found 1 issue, already fixed in 407ad78:

1. \`_github_error_message()\`'s output was passed unsanitized into \`_mark_failed\`/\`_requeue_for_retry\`, bypassing \`sanitize_error()\` — the redaction/truncation guard every other error write-path in this file uses (added in \`4826a2d\` for issue #86, since \`jobs.result\` is exposed verbatim via \`GET /jobs\` to workspace admins). This reintroduced the exact risk that fix closed: no length cap and no token-shaped-substring redaction on GitHub's raw error message before persisting it.

https://github.com/nazarli-shabnam/clevis/blob/6a869bb1320dc0daba61b23a8bfb87ac833fa86e/apps/worker/src/worker.py#L204-L212

Fixed by wrapping both call sites in \`sanitize_error(...)\`, matching the existing pattern at lines 152, 173, 180, 200 of the same file. Worker unit tests (13 non-DB-dependent tests in \`test_process_job.py\`) still pass after the fix.

PR #303, opened 2026-07-25
refactor(ui): drop the Jobs page, surface job status inline on Audit Log

### LINKED ISSUE #281: Jobs page and Audit Log page overlap in the UI — consolidate navigation around the cache-clear flow
Investigated the suspicion that Jobs and Audit Log show duplicate data (raised alongside #274). **Correcting the premise up front: they are not duplicate data.** The two tables share no columns — `jobs` has no `target` field, `audit_logs` has no `status` field.

- `jobs` (`apps/api/src/core/db.py:73-94`) has exactly **one** writer in the entire codebase: `cache_service.clear` (`apps/api/src/services/cache_service.py:19`), job_type `github.clear_actions_cache`. It tracks only that operation's async lifecycle (queued → processing → done/failed + result).
- `audit_logs` (`apps/api/src/core/db.py:62-70`) is a much broader append-only trail, written from 7 call sites across 5 files (installs, token resolve, automation dispatch, cache clear, etc.), with no status/lifecycle concept.

The only action that touches both is cache-clear: it writes a `jobs` row (live async status) and an `audit_logs` row whose payload embeds the `job_id` — a deliberate cross-reference, not duplicate reporting.

**The real problem is navigational, not data-model:** the Jobs page (`apps/ui/app/jobs/page.tsx`) is a full top-level nav destination dedicated to the status of exactly one background operation. A user trying to understand "what happened with my cache clear" (the same operation flagged as buried/confusing in #274) has to check two separate pages today — Jobs for live status, Audit for who/when — which is where the "am I looking at the same thing twice" feeling comes from, even though the underlying rows aren't actually redundant.

**Proposed direction** (not prescriptive — pick during implementation):
- (a) Surface job status inline on the relevant Audit Log row when a `job_id` is present in its payload (expandable or linked), and drop Jobs as a separate top-level nav item now that it's single-purpose; or
- (b) Keep both pages but cross-link them (Audit row links to its Job, Jobs row links to its Audit entry).

(a) is likely simpler given `jobs` currently has only one producer — but if `jobs` ever gains more job types in the future, that calculus could change, so worth confirming intent before committing to removing the page.

Related: #274 (the cache-clear operation itself is buried and its clear requests currently fail due to a missing `key` param) — that's about the operation being hard to use; this issue is about reconciling its status across two disconnected pages afterward.

No backend/schema change implied — this is a UI/IA consolidation only.

nazarli-shabnam:
## Summary

Fixes #281.

Investigation confirmed Jobs and Audit Log are not duplicate data: `jobs` has exactly one writer (`cache_service.clear`) and no `target` column; `audit_logs` has no `status` column. The only overlap is the cache-clear flow, which writes both — a `jobs` row for live async status, and an `audit_logs` row whose payload embeds that job's `id` (a deliberate cross-reference). The actual problem was navigational: understanding "what happened with my cache clear" meant checking two disconnected pages.

Implements the confirmed direction (a):
- The Audit Log page now also fetches the existing `GET /jobs` list (same data source the Jobs page polled — no new backend endpoint) and, for any row whose `payload` embeds a `job_id`, renders that job's live status inline in a new "Job status" column.
- Drops `apps/ui/app/jobs/page.tsx`, its `layout.tsx`, its sidebar nav entry, and its test file — coverage for that behavior now lives in `audit-page.test.tsx`.
- `cache-panel.tsx`'s post-clear "View in Job Queue" link now points at `/audit?job_id=...`; the Audit page reimplements the same scroll-to-and-highlight-the-matching-row behavior the Jobs page had for `?id=...`.
- Fixes the pre-existing `ACTION_TYPES` filter gap noted in the issue (missing `"cache.clear.queued"`) since it's directly relevant to finding the rows this change surfaces job status on — flagging it explicitly here per AGENTS.md rather than folding it in silently.

No backend/schema change — this is a UI/IA consolidation only.

## Test plan
- [x] `bun run typecheck`, `bun run lint`, `bun run build`, full `bunx vitest run` (352 tests) all pass locally
- [x] `diff-cover` locally: 100% on the sole changed file requiring new coverage (`apps/ui/app/audit/page.tsx`)
- [x] New `audit-page.test.tsx` cases: renders a row's live job status when its payload embeds a matching `job_id`; shows a dash for rows with no `job_id` or an unmatched one; doesn't crash on a malformed (non-JSON) payload; highlights the row matching a `?job_id=` deep link
- [x] Removed `jobs-page.test.tsx` (page deleted); updated `app-sidebar.test.tsx` to drop "Job Queue" from the shipped-nav-items list

## Summary by CodeRabbit

* **New Features**
  * Added live job status information to audit log entries.
  * Added deep links that highlight and scroll to a specific audit entry.
  * Added handling for queued cache-clear actions in audit filters.
  * Cache-clear results now link directly to the relevant audit log entry.

* **Changes**
  * Removed the standalone Job Queue page and its sidebar navigation link.


nazarli-shabnam:
### Code review

Found 2 issues, both fixed in b626bf6:

1. Three call sites still pointed at the deleted `/jobs` route after this PR removed it, including a live dead link a user could click into a 404 (AGENTS.md's review guidance for this PR explicitly called out confirming no dangling references to the removed page).

https://github.com/nazarli-shabnam/clevis/blob/559a0d10e351a71750927838452291c01cb91dfe/apps/ui/components/activity-list.tsx#L72-L77

Also `apps/ui/components/breadcrumb.tsx` (dead `ROUTE_LABELS["/jobs"]` entry) and `apps/ui/e2e/auth.spec.ts` (mid-session-401 test navigated to `/jobs` specifically to trigger the 401 it asserts on).

2. The new per-row "Job status" column's `jobs` query had no `isError` handling, silently rendering the same "—" placeholder on a genuine fetch failure as it does for rows with no `job_id` — reintroducing the exact silent-error anti-pattern commit `a1129a1` fixed for this page's main query.

https://github.com/nazarli-shabnam/clevis/blob/559a0d10e351a71750927838452291c01cb91dfe/apps/ui/app/audit/page.tsx#L52-L59

PR #305, opened 2026-07-25
fix(ui): show a 'no account selected' prompt on the Repos page

### LINKED ISSUE #277: Repos page has no "no account selected" empty state
Six other pages (`releases`, home `page.tsx`, `activity`, `my/issues`, `my/prs`, `my/reviews`) all render the same "No account selected yet — pick an organization…" card when there's no active scope selected.

`apps/ui/app/repos/page.tsx` has no equivalent guard. Needs verifying exactly what it currently renders with no active scope (likely a broken or blank table rather than a friendly empty state) and bringing it in line with the pattern used on the other pages.

nazarli-shabnam:
## Summary

Fixes #277.

The six other pages that render `useActiveScope()`-driven content (Releases, Overview, Activity, My Issues, My PRs, My Reviews) all show the shared `EmptyStateNoAccount` card when nothing is selected in the profile menu. The Repos page had no equivalent — with no active scope, it silently rendered just a blank org/token form and disabled "Load repositories" button, with no explanation.

Unlike those six pages, Repos' org field is a manual search tool that already works independent of the active scope — it only pre-fills the org input when an org-kind scope is active, but stays fully usable for any org the user types regardless of scope. So rather than hard-gating the whole page behind the scope check (which would remove the ability to manually search any org), the `EmptyStateNoAccount` banner renders as an informational addition above the existing form when there's no scope, matching the visual pattern used everywhere else without changing the page's actual (already-correct) behavior.

## Test plan
- [x] `bun run typecheck`, `bun run lint`, full `bunx vitest run` (379 tests) all pass locally
- [x] `diff-cover` locally: 100% on the changed lines in `apps/ui/app/repos/page.tsx`
- [x] New tests in `repos-page.test.tsx`: shows the prompt with no active scope (without blocking the org field); prompt disappears once an org scope is set

## Summary by CodeRabbit

* **New Features**
  * Added an informational message when no account is selected on the repositories page.
  * The message appears after the active account status is checked and does not interfere with manual organization selection.
  * When an account is selected, its organization is displayed and the no-account message remains hidden.

### DIFF
--- a/apps/ui/app/repos/page.tsx
+++ b/apps/ui/app/repos/page.tsx
@@ -4,6 +4,7 @@ import { useEffect, useState } from "react"
 import Link from "next/link"
 import { useMutation, useQuery } from "@tanstack/react-query"
 import { PageHeader } from "@/components/page-header"
+import { EmptyStateNoAccount } from "@/components/empty-state"
 import { Input } from "@/components/ui/input"
 import { Button } from "@/components/ui/button"
 import { Skeleton } from "@/components/ui/skeleton"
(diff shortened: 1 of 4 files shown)


nazarli-shabnam:
### Code review

Found 1 issue, fixed in e9314b6:

1. The default `EmptyStateNoAccount` copy ("this page has nothing to query") is inaccurate for the Repos page — its org field is a manual search that works independent of the active scope, as this PR's own commit message and tests state. Shipping the unmodified banner would show a message that contradicts the fully-usable form directly below it.

https://github.com/nazarli-shabnam/clevis/blob/1b69e93e5e74fdd644b9b4eb048ea7e97f6f0554/apps/ui/app/repos/page.tsx#L270-L280

Fixed by adding an optional `message` override prop to `EmptyStateNoAccount` (`apps/ui/components/empty-state.tsx`) and using it here with accurate copy.

PR #306, opened 2026-07-25
fix(ui): trim owner before matching personal installations

nazarli-shabnam:
## Summary

CodeRabbit flagged this on PR #299 (\`hasInstallationForOwner\` comparing \`account_login\` against untrimmed owner input on Automation/Security/Repos, so \`"acme "\` never matched an \`"acme"\` installation and incorrectly left the token field visible), and I fixed it there — but PR #299 got merged before that follow-up commit was pushed, so both the fix and its regression tests never actually landed on \`main\`. This PR restores them.

- \`apps/ui/app/automation/page.tsx\`, \`apps/ui/app/security/page.tsx\`, \`apps/ui/app/repos/page.tsx\`: \`installs.some((i) => i.account_login === owner)\` → \`=== owner.trim()\`, matching each page's already-trimmed org-scoped installations query.
- \`apps/ui/components/repo/cache-panel.tsx\` is unaffected — its \`owner\` is a route-derived prop (already clean), not free-text state, and its own installations check already uses \`owner\` consistently throughout.

No backend/schema change.

## Test plan
- [x] \`bun run typecheck\`, \`bun run lint\`, full \`bunx vitest run\` (all pass locally)
- [x] \`diff-cover\` locally: 100% on the three changed lines
- [x] Restored the three regression tests (one per page) asserting the token field still hides when the entered org has trailing whitespace but a matching installation exists

### DIFF
--- a/apps/ui/app/automation/page.tsx
+++ b/apps/ui/app/automation/page.tsx
@@ -60,7 +60,7 @@ export default function AutomationPage() {
     retry: false,
   })
   const hasInstallationForOwner =
-    installs.some((i) => i.account_login === owner) || (orgInstallsQuery.data?.length ?? 0) > 0
+    installs.some((i) => i.account_login === owner.trim()) || (orgInstallsQuery.data?.length ?? 0) > 0
 
   const resolveMutation = useMutation({
     mutationFn: (org: string) => api.tokens.resolve(org),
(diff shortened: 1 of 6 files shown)


nazarli-shabnam:
### Code review

No issues found. Confirmed this fully addresses CodeRabbit's original comment on PR #299 (the `.trim()` fix and regression tests for all three named pages), checked for consistency with existing `owner.trim()` usage elsewhere in each file, and reviewed git history and code comments — nothing contradicted.

PR #320, opened 2026-08-12
feat(db): add tenants/memberships schema and org-tenant backfill

nazarli-shabnam:
## Summary

Issue #190 (S2 multi-tenancy), PR 2 of the 7-PR breakdown recorded on the issue: adds the `tenants` and `memberships` tables and backfills one row per existing `org`/`org_membership`. Pure schema addition — nothing reads from these tables yet; RLS, dual-write, and cutover are later PRs in the same sequence.

**Base branch note:** targets `190-worker-db-credential-separation` (PR #314), not `main` — this PR's migrations start at `0021`, which requires #314's `0020_grant_worker_role_privileges` to exist first (see the renumbering comment on #190: https://github.com/nazarli-shabnam/clevis/issues/190#issuecomment-5265790097). Retarget to `main` once #314 merges.

**Unrelated commit included:** also carries a cherry-picked test fix (`fix(ui): stop overview-page milestone burndown test date from going stale`, also opened standalone as #319 against `main`) — it was needed to get the pre-push hook green on this branch. Will be a no-op once #319 merges and this branch rebases past it.

## Schema change (per AGENTS.md's migration guardrail)

This PR includes a schema change: migrations `0021_add_tenants_and_memberships` and `0022_backfill_org_tenants`.

- `0021`: creates `tenants` (kind='org'|'personal', 1:1 to an org or user via partial unique indexes, `ck_tenants_kind_xor` check constraint) and `memberships` (tenant_id + user_id + role, unique per tenant/user). Purely additive — no existing table touched.
- `0022`: backfills one `tenants` row per existing `org` and one `memberships` row per existing `org_membership`, carrying the same role. Zero ambiguity — every org gets exactly one tenant.

Both verified against a real Postgres instance: full `alembic upgrade head` / `downgrade` cycle, including a seeded org+membership row to exercise the backfill logic (confirmed the correct tenant/membership row was produced, then cleaned up). `pytest -q` — 606 passed.

## What changed

- `apps/api/alembic/versions/0021_add_tenants_and_memberships.py` (new)
- `apps/api/alembic/versions/0022_backfill_org_tenants.py` (new)
- `apps/api/src/core/db.py`: adds `Tenant` and `Membership` SQLAlchemy models matching the migration schema.

## Test plan

- [x] `alembic upgrade head` / `downgrade` cycle against real Postgres, with seeded org/membership data to exercise the backfill.
- [x] `pytest -q` — 606 passed.
[...]

Issue #190, opened 2026-07-17
[Not ready] feat: S2 Multi-tenancy (tenants/memberships/RLS) not started


nazarli-shabnam:
> **⚠️ NOT PLANNED FOR NEAR-TERM IMPLEMENTATION.** This is a large architectural pivot, not incremental feature work. Filed for tracking/visibility only — do not pick this up without an explicit decision to begin the SaaS migration.

## Problem

`docs/plan.md`'s migration step **S2 — Multi-tenancy** is marked `[ ]` (not started). The entire codebase today is single-org: there is no `tenants` or `memberships` table, no `tenant_id` column on any of the existing tables (`github_installations`, `audit_logs`, `jobs`, `saved_tokens`, `users`, `orgs`, `org_memberships`, `invitations`, `app_config` — see `apps/api/src/core/db.py`), and no Postgres Row-Level Security policies.

## What this actually is

Per `docs/architecture/clevis-architecture.md`, S2 converts Clevis from "one Postgres database, implicitly single-tenant" into a real multi-tenant SaaS: every table gains `tenant_id`, RLS policies enforce isolation at the database layer (not just in application code), and the existing `Org`/`OrgMembership` model is extended into a full `tenants`+`memberships` model with owner/member roles resolved per-tenant.

## What needs to land first

- A concrete decision that Clevis is moving to multi-tenant SaaS (this is a business/product decision, not just an engineering one).
- S1 fully complete (GitHub App auth is live per #140/#56, but see the companion "legacy PAT still required" issue — Security/Cache pages still use PAT fallback, which needs to be gone before tenant isolation matters for those code paths).
- A migration strategy for backfilling `tenant_id` on all existing rows without downtime or data loss — this touches every table in the schema and is exactly the kind of migration `AGENTS.md`'s "Guardrails for AI Agents" section calls out as needing deliberate, non-rushed care.
- Decision on RLS enforcement strategy (session-variable-based tenant context vs. app-level filtering as a stopgap).

## Severity

N/A — tracking only, not scoped for implementation.

nazarli-shabnam:
## Design decision: proceeding with the original tenants + memberships + RLS design

[...]

PR #323, opened 2026-08-13
feat(db): land tenants/memberships schema + dual-write into main

nazarli-shabnam:
## Summary

Issue #190 (S2 multi-tenancy) — lands PRs 2, 3, and 4 of the 7-PR breakdown into `main` in one shot: `tenants`/`memberships` schema + org-tenant backfill, personal-tenant backfill + remaining `tenant_id` columns, and application-code dual-write. These were already reviewed and merged individually (#320, #321, #322) but landed on a chain of intermediate staging branches rather than `main` directly — this PR is the one that actually brings that work into `main`. No new work beyond what those three PRs already contained; this is purely a consolidation merge.

**Why a single consolidating PR instead of three separate ones targeting `main`:** #320 merged into `190-worker-db-credential-separation`, #321 merged into a *different* branch (`190-tenants-memberships-schema`), and #322 merged into yet another (`190-personal-tenants-remaining-columns`) — each was correctly reviewed and merged into its own stated base, but merges don't propagate backward between sibling branches, so no single one of those branches ended up with the full stack except this one (`190-personal-tenants-remaining-columns`, which is `git diff --stat`-confirmed to contain exactly PR2+PR3+PR4's combined 23 files / 997 insertions and nothing else relative to `main`).

## Schema change (per AGENTS.md's migration guardrail)

**Yes, this PR includes schema changes** — migrations `0021` through `0028` (all already reviewed individually in #320/#321):

- `0021_add_tenants_and_memberships` / `0022_backfill_org_tenants` — new `tenants`/`memberships` tables + one row per existing org/org_membership.
- `0023_backfill_personal_tenants` — one personal tenant/membership per existing user.
- `0024_add_orgs_tenant_id` / `0025_add_github_installations_tenant_id` / `0026_add_invitations_tenant_id` — add + backfill, all **nullable** (deliberately — see design-bug note in #321's own description; app code didn't populate these until PR 4).
- `0027_add_scan_results_and_saved_tokens_tenant_id` — add + best-effort backfill, stays nullable (some legacy rows unmatched, documented not chased).
- `0028_add_audit_logs_tenant_id` — add nullable, no backfill (free-text `actor`/`target`, not FK-joinable).

No `NOT NULL` enforcement happens in this PR — that's the next PR (5a) in the sequence, now that dual-write (PR 4, included here) has landed and there's app code actually populating these columns going forward.

## What changed

Application code (PR 4, new since #321's description): `apps/api/src/repositories/tenant_repo.py` (new) centralizes all `Tenant`/`Membership` writes; wired into `org_repo.get_or_create`, `org_membership_repo.get_or_create`/`update_role`/`delete`, `installation_repo.upsert`, `invitation_repo.create`, and all 3 `User`-creation sites (`auth.py` setup/register, `github_auth.py` find_or_create_user) — every write to `orgs`/`org_memberships`/`users`/`github_installations`/`invitations` now also writes the matching `tenants`/`memberships` row. No read-path changes.

Full file list and individual rationale already documented on #320 and #321's PR descriptions (both merged, still viewable).

## Test plan

- [x] `pytest -q` — 633 passed on the source branch before this PR was opened (up from `main`'s current baseline).
- [x] Each of #320/#321/#322 was independently verified against real Postgres (migration upgrade/downgrade cycles, seeded backfill data) and had full pre-commit/pre-push hooks pass before merging.
- [x] `git diff --stat origin/main..190-personal-tenants-remaining-columns` confirmed this PR's diff is exactly the union of #320+#321+#322's changes, nothing more.

## Summary by CodeRabbit

* **New Features**
  * Added tenant-based ownership for organizations and personal accounts.
  * New users automatically receive a personal tenant with administrator access.
  * GitHub installations, invitations, scan results, saved tokens, and audit logs can now be associated with a tenant.
  * Organization membership changes stay synchronized with tenant memberships.
  * Added tenant-aware authorization for organization and personal account access.

* **Bug Fixes**
  * Existing records are automatically linked to the appropriate tenant where possible.
  * Prevented duplicate personal tenants and memberships during repeated operations.
  * Improved cleanup of tenant and user access context between requests.

### DIFF
--- a/apps/api/alembic/versions/0021_add_tenants_and_memberships.py
+++ b/apps/api/alembic/versions/0021_add_tenants_and_memberships.py
@@ -53,6 +53,11 @@ def upgrade() -> None:
         unique=True,
         postgresql_where=sa.text("personal_user_id IS NOT NULL"),
     )
+    # (id, org_id) as a composite unique constraint -- redundant with id's own PK
+    # uniqueness on its own, but required so a later composite FK from orgs.tenant_id
+    # can reference this exact column pair (see migration 0024's reciprocal-association
+    # fix on orgs.tenant_id).
+    op.create_unique_constraint("uq_tenants_id_org_id", "tenants", ["id", "org_id"])
 
     op.create_table(
         "memberships",
@@ -67,6 +72,7 @@ def upgrade() -> None:
 
 def downgrade() -> None:
     op.drop_table("memberships")
+    op.drop_constraint("uq_tenants_id_org_id", "tenants", type_="unique")
     op.drop_index("uq_tenants_personal_user_id", table_name="tenants")
     op.drop_index("uq_tenants_org_id", table_name="tenants")
     op.drop_table("tenants")
--- a/apps/api/alembic/versions/0021_add_tenants_and_memberships.py
+++ b/apps/api/alembic/versions/0021_add_tenants_and_memberships.py
@@ -1,0 +1,72 @@
+"""Add tenants and memberships tables (issue #190, S2 multi-tenancy PR 2 of 7).
+
+Pure additive schema change, no backfill and no existing table touched --
+zero data-loss risk. See the design comment on #190 for the full plan this
+is one step of: https://github.com/nazarli-shabnam/clevis/issues/190
+
+`tenants` is a new table, not a rename of `orgs` -- every org gets a 1:1
+`tenants` row (kind='org') and every user gets an implicit personal tenant
+(kind='personal', added in a later PR), so personal-scope resources get
+real DB-level isolation once Row-Level Security lands, not just
+`owner_user_id` filtering. `memberships` mirrors `org_memberships`'
+role vocabulary ("admin"|"member") to avoid renaming that concept mid-flight.
+
+Revision ID: 0021
+Revises: 0020
+Create Date: 2026-08-12
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "0021"
+down_revision = "0020"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "tenants",
+        sa.Column("id", sa.Integer(), primary_key=True, autoincrement=True),
+        sa.Column("kind", sa.Text(), nullable=False),  # 'org' | 'personal'
+        sa.Column("org_id", sa.Integer(), sa.ForeignKey("orgs.id"), nullable=True),
+        sa.Column("personal_user_id", sa.Integer(), sa.ForeignKey("users.id"), nullable=True),
+        sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=False),
+        sa.CheckConstraint(
+            "(kind = 'org' AND org_id IS NOT NULL AND personal_user_id IS NULL) OR "
+            "(kind = 'personal' AND org_id IS NULL AND personal_user_id IS NOT NULL)",
+            name="ck_tenants_kind_xor",
+        ),
+    )
+    op.create_index(
+        "uq_tenants_org_id",
+        "tenants",
+        ["org_id"],
+        unique=True,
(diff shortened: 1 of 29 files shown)


nazarli-shabnam:
Ran `/code-review medium` against this PR's actual diff (`gh pr diff 323`, 24 files). Three findings, verified individually against the real diff and current code — none are merge-blockers, but flagging all three for transparency:

1. **`org_membership_repo.py` — `get_or_create`'s early-return path skips the `Membership` mirror write** (confirmed real, present in this PR's diff). When an `OrgMembership` row already exists, the function returns immediately without ever syncing to `tenants`/`memberships`. **This is already fixed** in the next PR in the stack (#324, `190-rls-not-null-rbac`, branched from this PR's head) — that fix isn't in *this* PR's diff since it was found and fixed afterward. **Real-world risk right now is nil**: verified nothing in `apps/api/src` reads the `memberships` table yet (`grep -rn "query(Membership)" apps/api/src/routers apps/api/src/services` → no matches) — the dual-write is write-only scaffolding until PR 6 (cutover) lands, so a temporarily-incomplete mirror has no behavioral effect today. Not blocking, but worth knowing this PR ships with that specific gap until #324 also reaches `main`.
2. **`tenant_repo.py` — three get-or-create functions hand-roll the same optimistic-insert/`IntegrityError`-retry pattern.** Legitimate DRY observation (matches a shape already used 2 more times elsewhere in the codebase, predating this PR), not a bug. Worth a follow-up refactor, not this PR's job.
3. **`scan_results_repo.py`/`tokens.py` never set `tenant_id` on new rows.** True, but this file isn't actually touched by this PR's diff (confirmed: `gh pr diff 323 | grep scan_results_repo` → no matches) — it's a pre-existing, already-documented scope decision (see this PR's own description: "`ScanResult`/`SavedToken` writes explicitly out of scope"), not a new issue introduced here.

No blocking issues found. Ready to merge whenever you are.

nazarli-shabnam:
## CodeRabbit + code-review findings fixed, re-verified

**CodeRabbit finding 1 (org_membership_repo.py, Major):** get_or_create/update_role didn't sync the tenant Membership mirror on every path.
-> Already fixed on #324 (which branches from this PR's branch) via _sync_membership_mirror. This PR's own diff doesn't need a separate fix -- it resolves automatically once #324 merges into 190-personal-tenants-remaining-columns. Merge #324 first, then this PR, so main ends up with the fix included. The CodeRabbit thread on this repo intentionally stays open until then; I verified directly that #324 already contains the fix rather than relying on that thread's status.

**CodeRabbit finding 2 (tenant_repo.py/auth.py/github_auth.py, Major):** ensure_personal_tenant committed the tenant+membership separately from the caller's User commit in all three registration flows -- a failure between the two commits could leave a User row with no personal tenant.
-> Fixed: added ensure_personal_tenant(..., commit=False) so the tenant/membership inserts land in the same transaction as the just-flushed User row, committed once. Verified with a regression test that simulates a mid-flow failure and confirms the User row rolls back too (test_ensure_personal_tenant_commit_false_is_atomic_with_the_caller).

**Follow-up /code-review medium finding:** the atomicity fix moved ensure_personal_tenant earlier in register(), widening a pre-existing narrow window where the verification email could be sent before the transaction durably commits.
-> Fixed: split verification-token generation (pre-commit) from the actual email send (post-commit), so the real network send only happens once the commit has succeeded. Applied to both register() and resend-verification.

**Follow-up finding:** the commit=False branch of ensure_personal_tenant inserted a Membership row unconditionally instead of checking first -- safe today (always a brand-new user_id) but fragile for a future caller.
-> Hardened to query-first, matching the committing branch's idempotent behavior.

All 635 backend tests pass, and CI is fully green (one earlier Docker/UI-build run hit the known transient Google-Fonts fetch failure -- unrelated to this backend-only diff -- and passed clean on rerun).

No blocking issues remain on this PR's own diff. Ready to merge once #324 lands first, per the merge-order note above.

PR #324, opened 2026-08-13
feat(db): enforce tenant_id NOT NULL + wire tenant session context (RBAC)

nazarli-shabnam:
## Summary

Issue #190 (S2 multi-tenancy), PR 5a of 7 (RLS enablement, split in two for risk containment — see design rationale below). Two things:

1. Enforces `NOT NULL` on `invitations.tenant_id` and `github_installations.tenant_id`, now that PR 4's dual-write (already merged) covers every write path.
2. Wires `require_org_role` and a new `require_personal_tenant` dependency to set `app.tenant_id`/`app.user_id` Postgres session variables — scaffolding for PR 5b's RLS policies, no behavior change yet (nothing reads these variables until PR 5b).

## Schema change (per AGENTS.md's migration guardrail)

**Yes** — migration `0029_enforce_tenant_id_not_null`. Re-runs the original backfill `UPDATE` from migrations `0025`/`0026` (idempotent) to catch any row created in the deploy gap between those migrations landing and PR 4's dual-write code actually shipping, then `ALTER COLUMN ... SET NOT NULL`.

**Data-loss/backfill risk, stated explicitly per the guardrail:** if a real gap exists beyond what the backfill can resolve, the `ALTER COLUMN` fails atomically and the whole migration rolls back — verified this directly against real Postgres: seeded a genuine unbackfillable gap (an org with no tenant row, an invitation pointing at it with `tenant_id=NULL`), confirmed `alembic upgrade head` failed with a clean `NotNullViolation` and rolled back to the prior revision with no partial state, then fixed the gap and confirmed a clean upgrade.

**`orgs.tenant_id` is deliberately excluded and stays nullable permanently** — a real structural finding from this PR: creating a brand-new org requires inserting both the `Org` row (needs `tenant_id`) and its reciprocal `Tenant` row (needs `org_id`, via the composite FK from migration `0024`) — each references the other's not-yet-existing id. Postgres `NOT NULL` constraints can't be deferred (only FK/UNIQUE/PK/EXCLUSION constraints can), so there's no way to insert a genuinely new org+tenant pair in one transaction under a hard `NOT NULL`, short of a materially more complex deferred-FK migration. Decided with the repo owner to keep `orgs.tenant_id` nullable and rely on `org_repo.get_or_create`'s existing self-healing dual-write instead — same accepted pattern already used for `scan_results`/`saved_tokens.tenant_id`.

## What changed

- `apps/api/alembic/versions/0029_enforce_tenant_id_not_null.py` (new) — includes the verification-query runbook inline in its docstring (a standalone `docs/` file was considered but this repo's `.gitignore` reserves `docs/` for local Claude planning artifacts, only `docs/self-hosting.md` is actually tracked).
- `apps/api/src/core/db.py` — `Invitation.tenant_id`/`GitHubInstallation.tenant_id` become non-optional `Mapped[int]`; `Org.tenant_id` stays `Mapped[int | None]` with an updated docstring explaining why.
- `apps/api/src/core/rbac.py` — `require_org_role` now sets `app.tenant_id`/`app.user_id` (plain `SET`, not `SET LOCAL` — a request's `Session` lifetime doesn't cleanly map to one transaction) after resolving org+membership, self-healing `org.tenant_id` inline if a legacy row resolved via `get_by_login` never went through `get_or_create`'s dual-write. New `require_personal_tenant` dependency does the same for the caller's own personal tenant. Not yet wired into any `/me/...` route — those that go through `token_resolution.resolve_owner_token` can resolve to either an org or personal tenant depending on the `owner` path param, which is a real design fork deferred to before PR 6, not decided here.
- `apps/api/src/repositories/tenant_repo.py` — new `upsert_membership` (get-or-create plus role reconciliation).
- `apps/api/src/repositories/org_membership_repo.py` — **CodeRabbit finding on PR #323, fixed:** `get_or_create`'s early-return path (membership already exists — the common case, since `org_provisioning.py`'s reconcile loop calls this on every login) and its concurrent-insert-race recovery path both skipped the `Membership` mirror write entirely; `update_role`'s mirror call also silently no-op'd if the mirror row didn't exist yet (`tenant_repo.update_membership_role` returns `None` on a missing row and nothing checked that). All three functions now funnel through a shared `_sync_membership_mirror` helper backed by the new `upsert_membership`, so the mirror self-heals on every call regardless of which path is taken.

**Also evaluated, deliberately not fixed here:** a second CodeRabbit finding on PR #323 about `tenant_repo.ensure_personal_tenant` and its 3 callers (`auth.py` setup/register, `github_auth.py`) not being fully atomic with the user-row commit — a real gap (a failure between the separate commits could leave an orphaned `User` with no personal tenant), but CodeRabbit itself tagged it "Heavy lift," and the existing codebase already has the same multi-commit-within-one-request pattern elsewhere (e.g. `org_repo.get_or_create`'s own two-phase org-insert-then-tenant-link). Fixing it properly would mean restructuring commit boundaries across 3 files for a low-probability edge case, out of scope for this PR.

## Test plan

- [x] New `apps/api/tests/test_rbac.py` — asserts `require_org_role`/`require_personal_tenant` actually set the session variables (and don't on a 403), using `current_setting()`.
- [x] Extended `test_org_membership_repo.py` — two regression tests reproducing the CodeRabbit-flagged gap directly (a missing mirror row and a stale mirror role both get repaired on the next `get_or_create` call).
- [x] Migration verified against real Postgres: downgrade/upgrade cycle, a deliberately-seeded unbackfillable gap causing a clean atomic failure (see above), then a clean upgrade after fixing it.
- [x] `pytest -q` — 629 passed.
- [x] `bun run test` via the pre-push hook — 388 passed (no UI files touched by this PR; one earlier push attempt hit an unrelated pre-existing flaky font-import test under local resource contention, resolved on retry).
- [x] `ruff check` on all touched Python files — clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved tenant and user context isolation when database connections are reused.
  - Automatically repairs missing or outdated organization membership links.
  - Ensures required tenant associations are populated and consistently maintained.
  - Strengthened authorization handling for organization and personal-tenant access.

- **Data Integrity**
  - Tenant associations for invitations and GitHub installations are now required.

- **Tests**
  - Added coverage for context cleanup, authorization behavior, membership repair, and concurrent membership updates.

### DIFF
--- a/apps/api/alembic/versions/0029_enforce_tenant_id_not_null.py
+++ b/apps/api/alembic/versions/0029_enforce_tenant_id_not_null.py
@@ -1,0 +1,90 @@
+"""Enforce NOT NULL on invitations/github_installations.tenant_id (issue #190, PR 5a of 7).
+
+Migrations 0025/0026 added these columns nullable because, at the time, no
+application code populated tenant_id yet -- enforcing NOT NULL then would
+have broken installation/invitation creation the moment those migrations
+deployed. PR 4 (dual-write, already merged) closed that gap: every write
+path now sets tenant_id via src.repositories.tenant_repo. This migration
+re-runs each column's original backfill UPDATE (idempotent -- a no-op for
+already-tenanted rows) to catch any row created in the window between PR 3's
+migrations landing and PR 4's dual-write code actually deploying, then
+enforces NOT NULL.
+
+orgs.tenant_id is deliberately NOT included here and stays nullable
+permanently: creating a brand-new org requires inserting both the Org row
+(needs tenant_id) and its reciprocal Tenant row (needs org_id via
+fk_orgs_tenant_id_reciprocal's composite FK, migration 0024) -- each
+references the other's not-yet-existing id. Postgres NOT NULL constraints
+cannot be deferred (only FK/UNIQUE/PK/EXCLUSION constraints can), so there
+is no way to insert a genuinely new org+tenant pair in one transaction if
+orgs.tenant_id is hard NOT NULL, short of a more complex deferred-FK
+migration (an explicit tradeoff, not attempted here -- see the PR
+description). org_repo.get_or_create's self-healing dual-write plus the
+verification queries below are the ongoing guarantee instead, the same
+pattern already accepted for scan_results/saved_tokens.tenant_id.
+
+Data-loss/backfill risk (AGENTS.md guardrail): if a real gap exists beyond
+what the backfill UPDATE can resolve (e.g. an installation/invitation whose
+org's tenant row was never created, not just never linked), the
+ALTER COLUMN ... SET NOT NULL below fails atomically and the whole migration
+rolls back -- it does not silently succeed with missing data. Run the
+verification queries below against the target environment before applying
+this migration there, and investigate any nonzero count rather than
+retrying blindly. Verified end-to-end against real Postgres for this PR: a
+deliberately-seeded gap (an org with no tenant row, an invitation pointing
+at it) reproduced the atomic-rollback failure below before being fixed.
+
+    -- every org has a reciprocal tenant
+    SELECT count(*) FROM orgs o WHERE NOT EXISTS (
+        SELECT 1 FROM tenants t WHERE t.org_id = o.id AND t.kind = 'org');
+
+    -- invitations/github_installations.tenant_id are populated
+    SELECT count(*) FROM invitations WHERE tenant_id IS NULL;
+    SELECT count(*) FROM github_installations WHERE tenant_id IS NULL;
+
+    -- every org_membership has a matching Membership row (dual-write parity)
+    SELECT count(*) FROM org_memberships om JOIN orgs o ON o.id = om.org_id
+    WHERE o.tenant_id IS NOT NULL AND NOT EXISTS (
+        SELECT 1 FROM memberships m WHERE m.tenant_id = o.tenant_id
+        AND m.user_id = om.user_id AND m.role = om.role);
+
+Revision ID: 0029
+Revises: 0028
+Create Date: 2026-08-13
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "0029"
+down_revision = "0028"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+
+    conn.execute(
+        sa.text(
+            "UPDATE invitations i SET tenant_id = t.id "
+            "FROM tenants t WHERE t.org_id = i.org_id AND t.kind = 'org' AND i.tenant_id IS NULL"
+        )
+    )
+    op.alter_column("invitations", "tenant_id", nullable=False)
+
+    conn.execute(
+        sa.text(
+            "UPDATE github_installations gi SET tenant_id = t.id "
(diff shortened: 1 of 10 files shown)


nazarli-shabnam:
## /code-review medium — verified, no blockers

Checked out the exact PR head (`a860a2d`) and cross-checked every finding against `gh pr diff 324`'s actual file list before drawing any conclusion.

**CI:** All checks green. The one E2E failure that showed up was a Docker build step failing to fetch a Google Font (`fonts.gstatic.com`) — a network dependency, unrelated to this PR's diff (which only touches `apps/api/**`, no UI files). Reran the job with no code changes and it passed, confirming it was transient infra flakiness, not a regression.

**Review finding (non-blocking):** `get_db()`'s `finally` block (`apps/api/src/core/db.py:295`) unconditionally runs a rollback + two `RESET` statements + commit on every request through `Depends(get_db)`, even though `app.tenant_id`/`app.user_id` are only ever set by `require_org_role`/`require_personal_tenant`. The ~60 other routes that don't go through those dependencies pay two no-op round-trips per request. Correctness is unaffected — this is a latency nit, safe to defer (e.g. gate the reset on a flag set only when `_set_tenant_session_context` actually ran).

**No blocking issues found.** This covers PR 5a of the #190 breakdown: NOT NULL enforcement on `invitations.tenant_id`/`github_installations.tenant_id` (migration 0029), RBAC session-variable wiring, the connection-pool leak fix, and the `org_repo.ensure_tenant_linked`/`tenant_repo.upsert_membership` cleanups from the two prior review rounds. Ready to merge whenever you are.

nazarli-shabnam:
## Second round: CodeRabbit + code-review findings fixed, re-verified

Two more rounds of fixes landed on top of the prior "ready to merge" state, after you asked me to address CodeRabbit's comments on this PR:

**CodeRabbit finding (org_membership_repo.py, Major/Security):** a concurrent `delete()` could interleave between `get_or_create()`/`update_role()`'s membership lookup and the tenant-mirror sync, resurrecting the mirror right after revocation.
→ Fixed with `SELECT ... FOR UPDATE` locking on the lookup, and reordering `update_role` so the mirror sync commits before the lock releases. Verified with a real two-connection concurrency test (`test_update_role_blocks_a_concurrent_delete_until_the_mirror_sync_commits`) — confirmed it fails without the fix and passes with it.

**Follow-up `/code-review medium` finding:** the same class of race existed in `delete()` itself, from the opposite direction — its old two-phase commit released the lock before touching the mirror, so a concurrent `get_or_create()` could recreate a membership whose mirror then got deleted out from under it.
→ Fixed by reordering `delete()` to remove the mirror before committing. Verified with `test_delete_blocks_a_concurrent_get_or_create_until_the_mirror_delete_commits`, same fails-without/passes-with verification.

**CodeRabbit nitpick:** the connection-leak regression test only asserted `app.tenant_id` was reset, not `app.user_id`. Fixed.

All 646 backend tests pass, both new concurrency tests run stable across repeated runs, and CI is fully green (one E2E run hit the known transient Google-Fonts fetch failure in the Docker build — unrelated to this diff — and passed clean on rerun).

**No blocking issues remain. Ready to merge whenever you are.**

Issue #330, opened 2026-08-14
API DB role is a Postgres superuser -- RLS (migrations 0030/0031) enforces nothing in production


nazarli-shabnam:
## Problem

Issue #190's whole point was database-enforced (not just app-level) tenant isolation via Postgres Row-Level Security. Migrations `0030` and `0031` (issue #190, PRs #325/#328) add correct, verified RLS policies -- verified against a manually-created non-superuser test role -- but **the API's actual DB connection role is a Postgres superuser**, and superusers unconditionally bypass RLS, `ENABLE` or `FORCE`, no exceptions.

Confirmed directly against the running system:
```sql
SELECT rolname, rolsuper, rolbypassrls FROM pg_roles WHERE rolname = current_user;
-- clevis | t | t
```

`docker-compose.yml` sets `POSTGRES_USER: ${DB_USER}`, and the official `postgres` Docker image always makes that role the `initdb` bootstrap superuser -- there is no way to make it non-superuser through that env var alone.

**Net effect:** every RLS policy added for #190 is logically correct but has zero actual effect on the real deployed system today. Cross-tenant reads/writes are only prevented by application-level checks (`require_org_role`, `require_personal_tenant`, etc.), exactly as before RLS was added -- the "defense in depth" #190 set out to build doesn't currently exist.

## What needs to happen

Give the API a dedicated, non-superuser Postgres role, mirroring the pattern migration `0020` already built for the *worker* (`clevis_worker` role + explicit `GRANT`s, see `WORKER_DB_PASSWORD` in `AGENTS.md`'s Development setup section):

- A new role (e.g. `clevis_api`), `NOSUPERUSER`.
- A migration granting it exactly the privileges the API needs across all its tables (SELECT/INSERT/UPDATE/DELETE as appropriate per table) -- broader than the worker's grant since the API touches nearly every table.
- Updated connection string / entrypoint wiring so the API actually connects as this role instead of the bootstrap superuser, in both docker-compose and any other deployment path.
- Re-verification of every RLS scenario already tested against a throwaway test role (see PR #328's description) against this *real* role instead, since that's the actual enforcement boundary that matters.
- Docs update (`AGENTS.md`, `docs/self-hosting.md`) once landed.

## Why this matters

Without this, #190 is incomplete relative to its own stated goal even though every planned PR (1-6) has merged -- the schema/migrations/app-code work is all real and correct, but the actual security property (DB-enforced isolation surviving an application-layer bug) doesn't hold yet.

[...]

PR #332, opened 2026-08-14
feat(db): create clevis_api Postgres role and grant table privileges

nazarli-shabnam:
Issue #330: the API currently connects as the initdb bootstrap superuser (DB_USER), which unconditionally bypasses Row-Level Security regardless of ENABLE/FORCE -- migrations 0030/0031's RLS policies are logically correct but currently enforce nothing in production. This mirrors migration 0020's clevis_worker pattern: a new init script creates a clevis_api login role on a fresh data volume when API_DB_PASSWORD is set, and migration 0032 grants it SELECT/INSERT/UPDATE/DELETE on every table the API's models touch.

Purely additive -- nothing connects as clevis_api yet. That cutover (switching the API's runtime engine to actually use this role) is a deliberately separate follow-up PR, since it's the actual behavior change and needs its own focused review plus a bake period.

## Summary

Closes #

## Checks

- [ ] `cd apps/ui && bun run typecheck && bun run build` (if UI changed)
- [ ] `python -m compileall apps/api/src apps/worker/src` (if API/worker changed)
- [ ] Relevant `pytest` tests pass (if API/worker changed)
- [ ] Docker images still build, if `Dockerfile`/deps changed
- [ ] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)

## Summary by CodeRabbit

* **New Features**
  * Added optional configuration for a dedicated, non-superuser API database role.
  * API requests can use restricted database access while migrations retain required administrative access.
  * Improved tenant-aware access controls and audit tracking across organizations, personal accounts, tokens, installations, and scan results.
  * Existing deployments remain unchanged when the optional setting is omitted.
* **Documentation**
  * Added setup instructions for new and existing deployments.
  * Documented required permissions, migration considerations, and provisioning limitations for existing databases.


nazarli-shabnam:
@coderabbitai Addressed CodeRabbit's two findings directly in migration `0032` (folded in schema USAGE + per-sequence grants, since `0032` hadn't merged yet and wasn't "already applied" under AGENTS.md's edit-lock rule) plus a new atomic existing-deployment provisioning step in `docs/self-hosting.md`. Verified upgrade/downgrade against a real Postgres role directly (insert succeeds after upgrade, zero privileges after downgrade). The OpenGrep SQL-injection flag on the f-string-built `DO $$` blocks is a false positive — the interpolated values are a fixed internal table/sequence name list, never user input.

Issue #191, opened 2026-07-17
[Not ready] feat: S4 Event processors + normalized model + aggregates not started


nazarli-shabnam:
> **⚠️ NOT PLANNED FOR NEAR-TERM IMPLEMENTATION.** This is a large architectural pivot, not incremental feature work. Filed for tracking/visibility only — do not pick this up without an explicit decision to begin the SaaS migration.

## Problem

`docs/plan.md`'s migration step **S4 — Event processors + normalized model + aggregates** is marked `[ ]` (not started). There is no consumer fleet processing webhook events, no idempotent dedupe-by-delivery-id logic, and no materialized aggregate tables anywhere in the codebase — `apps/worker/src/worker.py` today is a single poller handling exactly one job type (`github.clear_actions_cache`, see #143), not a general event-processing pipeline.

## What this actually is

Per `docs/architecture/clevis-architecture.md`, S4 is the layer that turns raw webhook payloads (received by S3's receiver) into normalized, deduplicated database rows and pre-computed aggregate tables, so the API (S6) can serve reads without hitting GitHub at request time.

## What needs to land first

- **S3 (webhook ingestion + queue)** — partially tracked by #142 (missing webhook-receiving endpoint), but S3's full scope also includes a managed queue (Redis Streams or SQS) and raw-payload object storage, neither of which exist yet.
- A managed queue service (Redis Streams / SQS or equivalent) provisioned and wired into deployment — today's `jobs` table + `SELECT FOR UPDATE SKIP LOCKED` pattern is explicitly noted in the architecture doc as being kept only for user-triggered ops, not high-volume event processing.
- Object storage for raw webhook payloads.
- A normalized event schema design (not yet drafted anywhere in `docs/`).

## Severity

N/A — tracking only, not scoped for implementation.

nazarli-shabnam:
Design decision on the S2 prerequisite this issue depends on: after review, the team decided to build the *original* tenants + memberships + Postgres RLS design for S2 (despite `docs/plan.md`'s 2026-07-23 status note calling that design superseded by the lighter `orgs`/`org_memberships` model already shipped). Full design (schema, migration sequencing, RLS mechanics, rollout plan, PR breakdown) is recorded as a comment on #190: https://github.com/nazarli-shabnam/clevis/issues/190#issuecomment-5139514355

This issue remains blocked until #190 lands — still not scoped for implementation on its own.

nazarli-shabnam:
[...]

PR #337, opened 2026-08-16
fix(api): resolve installation tenant_id via SECURITY DEFINER for RLS-safe webhook deletes

nazarli-shabnam:
## Summary

Fixes a genuine, already-live production bug — not a refactor. Second PR in the S3 (webhook ingestion + queue, issue #191) stack; branches off and targets `s3-webhook-ingestion-infra` (PR #336), not `main`.

Migration 0031 put `github_installations` under `FORCE ROW LEVEL SECURITY` with policy `tenant_id = app.tenant_id OR owner_user_id = app.user_id`. The GitHub webhook receiver (`POST /webhooks/github`) is unauthenticated by design (HMAC-signature-verified, not login-gated) and never calls `require_org_role`/`require_personal_tenant` — the only places that set those session vars. Under the real runtime `clevis_api` role (non-superuser since #330), `_handle_installation_deleted`'s delete against `github_installations` has been evaluating the policy to `NULL OR NULL` — **zero rows visible, every time**. GitHub's `installation.deleted` webhook fires, the signature verifies, the handler runs, the `DELETE` silently affects 0 rows, no audit log entry is written, and the stale row (and its now-invalid `token_ref`) is never cleaned up.

## What changed

- **New migration `0035`**: adds `resolve_installation_tenant_id(installation_id)`, a narrow `SECURITY DEFINER` SQL function owned by the table-owning migration role (the DB_USER superuser — unconditionally bypasses RLS). Grants `EXECUTE` to `clevis_api` only, not `BYPASSRLS` on the role itself (which would neuter RLS for every query the API makes, not just this one pre-authentication lookup).
- **`installation_repo.delete_by_installation_id`**: now calls that function first to resolve `tenant_id`, then calls the new `set_session_tenant` helper (`src/core/db.py`) to set `app.tenant_id` for the rest of the request — so the actual `SELECT`/`DELETE` that follows is evaluated under a session that now legitimately satisfies the policy's tenant_id-equality branch. This isn't a second RLS bypass; the SECURITY DEFINER function only ever answers the one narrow "what tenant does this installation belong to" question.
- **New helper `set_session_tenant`** (`src/core/db.py`): sets `app.tenant_id` alone, for system code paths (like this one) that resolve a tenant without an acting user.

## Why a migration is included

Yes — `apps/api/alembic/versions/0035_installation_tenant_lookup_fn.py` adds one new Postgres function (`resolve_installation_tenant_id`) and grants `clevis_api` `EXECUTE` on it. Upgrade is purely additive (zero data-loss risk). Downgrade drops the function, which would silently reintroduce the bug this PR fixes — do not run it against a real environment without explicit confirmation.

## Verification

- Reproduced the bug locally against a real `clevis_api` role: wrote a regression test that explicitly resets `app.tenant_id`/`app.user_id` mid-test (to accurately model the webhook receiver's fresh-connection, no-session-context reality — the existing `test_webhooks.py` suite doesn't catch this, since its shared `db` fixture uses SAVEPOINT-based test transactions where an earlier plain `SET` in the same test leaks across what would be separate requests/connections in production).
  - Confirmed the test **fails** (`removed == 0`) against the pre-fix code.
  - Confirmed it **passes** with the fix.
- Full `pytest -q` suite (569 passed, 3 xfailed) green locally under the real `clevis_api` role.
- `alembic check` reports no model/migration drift.
- Verified migration 0035 downgrades and re-upgrades cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed GitHub installation deletion when no authenticated user or tenant session context is available.
  * Ensured installation records are removed correctly while preserving tenant isolation.
  * Improved tenant resolution for background and system-level operations.
  * Restricted tenant lookup access to authorized database operations.

* **Tests**
  * Added regression coverage for deletion and tenant resolution under restricted database access conditions.

### DIFF
--- a/apps/api/alembic/versions/0035_installation_tenant_lookup_fn.py
+++ b/apps/api/alembic/versions/0035_installation_tenant_lookup_fn.py
@@ -60,6 +60,12 @@ def upgrade() -> None:
         $$;
         """
     )
+    # Postgres grants EXECUTE on a new function to PUBLIC by default (unlike tables) --
+    # left alone, the GRANT below wouldn't actually narrow anything, since every role
+    # that can connect could already call this function and read any installation's
+    # tenant_id, defeating the "narrow, deliberately scoped" bypass this migration's
+    # docstring describes (CodeRabbit finding on PR #337).
+    op.execute("REVOKE EXECUTE ON FUNCTION resolve_installation_tenant_id(integer) FROM PUBLIC")
     op.execute(
         """
         DO $$
--- a/apps/api/alembic/versions/0035_installation_tenant_lookup_fn.py
+++ b/apps/api/alembic/versions/0035_installation_tenant_lookup_fn.py
@@ -1,0 +1,77 @@
+"""Add resolve_installation_tenant_id() SECURITY DEFINER function (issue #191, S3 PR 2 of 4).
+
+**Fixes a genuine, already-live production bug** -- not a refactor. Migration 0031 put
+`github_installations` under `FORCE ROW LEVEL SECURITY` with policy
+`USING (tenant_id = app.tenant_id OR owner_user_id = app.user_id)`. The GitHub webhook
+receiver (`src/routers/webhooks.py`, `POST /webhooks/github`) is unauthenticated by
+design -- it verifies the request via HMAC signature, not a login session -- so it never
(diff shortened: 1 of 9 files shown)


nazarli-shabnam:
@coderabbitai review

PR #435, opened 2026-09-16
fix(api): enforce org-admin check on dry-run cache clear

### LINKED ISSUE #416: fix(api): dry-run cache-clear skips the org-admin membership check
_Found by an audit sweep on `main` @ `31ff8ce` (2026-09-09). Low impact, but an inconsistency with the repo's own RBAC guardrail._

### What
`apps/api/src/routers/actions_cache.py` (~L64-75): the personal cache-clear endpoint calls `resolve_owner_token(..., min_role="admin")` only on the non-dry-run path. When `payload.dry_run` is true, that call is skipped entirely — there is **no** membership / role check — and `clear()` still writes a `cache.clear` audit-log row (scoped to the caller's personal tenant).

So any authenticated user can generate `cache.clear` dry-run audit entries for arbitrary `{owner}/{repo}` strings they have no relationship with. No GitHub call is made and the row lands in the caller's own tenant, so real impact is limited to audit-log noise / minor info-leak-shaped log lines — but it contradicts `AGENTS.md`'s guardrail that privileged cache actions are gated by `require_org_role("admin")`, and the non-dry-run path on the same endpoint.

### Suggested fix
Run the same membership/role resolution before the dry-run branch (resolve first, then branch on `dry_run` for whether to actually enqueue), or explicitly document that dry-run is intentionally unauthenticated-beyond-login and skip the audit write for it.

nazarli-shabnam:
## What

Fixes #416.

`POST /me/repos/{owner}/{repo}/actions-caches/clear` only called
`resolve_owner_token(..., min_role="admin")` on the non-dry-run branch. A
`dry_run=true` request skipped every membership/role check entirely and still
wrote a `cache.clear.dry_run` audit-log row — any authenticated user could
generate audit entries against `{owner}/{repo}` strings for an org they have
no admin (or any) relationship with. No GitHub call is made on that path, so
real impact was audit-log noise / a minor info-leak-shaped log line, but it
contradicted this repo's own guardrail (AGENTS.md: privileged cache actions
are gated by `require_org_role`-equivalent checks) and the non-dry-run branch
of the same endpoint.

## Why this shape

The dry-run path doesn't need an actual GitHub token (`cache_service.clear`
never touches `token` when `payload.dry_run` is true), so simply moving the
existing `resolve_owner_token(min_role="admin")` call above the `dry_run`
branch would have forced token resolution (and installation/PAT
availability) for a call that never uses one — a behavior regression for any
dry run against a personal (non-org) repo with no connected GitHub App.

Instead, `token_resolution.py` now exposes `check_owner_role(...)`: the
membership/role-gate half of `resolve_owner_token`, split out so it can run
before the `dry_run` branch without pulling in token minting.
`resolve_owner_token` calls it internally too, so the non-dry-run path's
behavior (and its existing test coverage) is unchanged.

## Testing

- Added `test_personal_clear_caches_dry_run_rejects_org_member` (403 for a
  plain org member dry-running against the org's repo) and
  `test_personal_clear_caches_dry_run_allows_org_admin` (200 for an org
  admin) to `apps/api/tests/test_actions_cache.py`.
- Existing `resolve_owner_token` role-gating tests in
  `test_token_resolution.py` pass unchanged.
- Full suite: `pytest -q` — 1195 passed, 3 skipped.
- `bun run check` (typecheck + lint + build) and `bun run test` (518 passed)
  clean via the pre-push hook.

No migrations, no RBAC-model change (same `InsufficientOrgRole` → 403
mapping), no sensitive-config touch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted cache-clearing requests to organization administrators, including dry-run requests.
  * Users without the required administrative role now receive a 403 response.
  * Administrators can continue to perform successful dry-run cache-clearing requests.

* **Tests**
  * Added coverage verifying access control for dry-run cache clearing.


nazarli-shabnam:
### Code review

No issues found. Checked for bugs and AGENTS.md compliance.

PR #436, opened 2026-09-16
fix(api): validate membership_reconcile_* config keys as integers

### LINKED ISSUE #415: fix(api): /config accepts non-numeric membership_reconcile_* values (missing from _INT_KEYS)
_Found by an audit sweep on `main` @ `31ff8ce` (2026-09-09). Validation oversight._

### What
`apps/api/src/routers/config.py` (~L19-25) validates numeric `app_config` keys against `_INT_KEYS = {worker_poll_seconds, gap_heal_poll_seconds, gap_heal_stale_hours, pr_nudge_stale_days, digest_poll_seconds}`. But `apps/api/src/core/app_config.py` `_ACCEPTED_KEYS` also includes `membership_reconcile_poll_seconds` and `membership_reconcile_stale_hours`, which are **not** in `_INT_KEYS` (nor `_BOOL_KEYS` / `_ENUM_KEYS`).

So `PUT /config/membership_reconcile_poll_seconds` with body `"abc"` passes validation and is persisted. `apps/api/src/services/membership_reconcile_loop.py` / `membership_reconcile_sweep.py` then silently falls back to its default each iteration with a `log.warning` — the bad value just sits in the table.

Same latent gap for any future numeric key added to `_ACCEPTED_KEYS` without also updating `_INT_KEYS`.

### Suggested fix
Add `membership_reconcile_poll_seconds` / `membership_reconcile_stale_hours` to `_INT_KEYS` (with sensible clamps matching the loop's own bounds). Consider deriving the validator sets from a single per-key schema so `_ACCEPTED_KEYS` and the type sets can't drift again.

nazarli-shabnam:
## What

Fixes #415.

`app_config._ACCEPTED_KEYS` includes `membership_reconcile_poll_seconds` and
`membership_reconcile_stale_hours`, but the `/config` router's `_INT_KEYS`
validation set didn't. `PUT /config/membership_reconcile_poll_seconds` with a
non-numeric body value (e.g. `"abc"`) passed validation and was persisted to
`app_config`. `membership_reconcile_loop.py` / `membership_reconcile_sweep.py`
then silently fell back to their defaults every iteration (`log.warning`
only) instead of the bad write failing loudly at the point someone actually
made the typo.

## Fix

Added the two keys to `_INT_KEYS`. Same generic `>= 1` validation as every
other `_INT_KEYS` entry (`gap_heal_poll_seconds` etc.) -- the per-key upper
clamps (`[60, 3600]` for poll seconds, `[1, 168]` for stale hours) already
live where they belong, at read time in `membership_reconcile_loop.py` /
`membership_reconcile_sweep.py._read_stale_hours`, matching how
`gap_heal_poll_seconds` / `gap_heal_stale_hours` are handled. Didn't invent a
new per-key clamp mechanism in the router since nothing else there has one
either -- kept the diff to the one line the bug is actually about.

## Testing

- Added `test_update_config_membership_reconcile_keys_are_int_validated`
  (parametrized over both keys) to `apps/api/tests/test_auth.py`, asserting
  422 for a non-numeric value.
- Full suite: `pytest -q` -- 1195 passed, 3 skipped.

No migrations, no RBAC change, no sensitive config touched.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation for membership reconciliation settings.
  - Values must now be whole numbers of at least 1.
  - Invalid or non-numeric values are rejected instead of being saved, helping prevent misconfigured reconciliation behavior.

- **Tests**
  - Added regression coverage to verify that invalid membership reconciliation settings return a validation error and are not persisted.


nazarli-shabnam:
### Code review

No issues found. Checked for bugs and AGENTS.md compliance.

PR #437, opened 2026-09-16
fix(api): stop silent config/credential fallbacks from hiding outages

### LINKED ISSUE #414: fix(api): silent config/credential fallbacks hide outages (app_config.get_config, token_resolution)
_Found by an audit sweep on `main` @ `31ff8ce` (2026-09-09). Silent-failure hardening._

Two catch-all fallbacks that turn an outage into a silent behavior change, plus one minor doc gap.

### 1. `app_config.get_config` swallows every exception → serves the code default
`apps/api/src/core/app_config.py:49-57` — `except Exception: logger.warning("app_config read failed …, using default")`. A DB blip or a missing `SELECT` grant on `app_config` (e.g. under a constrained `clevis_api`/`clevis_worker` role) silently reverts **all** runtime config to code defaults: `registration_enabled` back to `"true"` (`auth.py:263` — re-opens self-registration), `pr_nudge_mode`, `digest_cadence`, every poll interval. A transient failure shouldn't flip a security-posture setting open.

### 2. `token_resolution._from_installation` swallows all HTTP errors → silent PAT fallback
`apps/api/src/services/token_resolution.py:52-57` — `except (httpx.HTTPStatusError, httpx.RequestError): … return None`. A rotated / malformed / revoked `GITHUB_APP_PRIVATE_KEY` (401 from GitHub's installation-token endpoint), or a GitHub 5xx, is treated identically to "no installation exists" and falls through to a client-supplied PAT or a generic `NoGitHubTokenAvailable`. A globally broken App key degrades **every** request to "paste a PAT" with only a per-call `warning` and no health signal.

### 3. Minor: score-formula doc imprecision
`packages/checks/src/checks/runner.py:29-44` collapses per-check exceptions to `status="error"` (generic text; real cause only in `logger.exception`). `analytics_service.py:24-26` correctly counts `error` as failed, but `AGENTS.md:57` states the score as `100 - (failed_count / total_count * 100)` with no mention of the `error` / `not_applicable` handling.

### Suggested fix
1. Narrow `get_config`'s catch to the expected transient errors, and/or keep the last-known-good value in memory instead of falling to the code default; at minimum log at `error` and expose it in `/healthz`.
2. Distinguish "no installation" (return `None`) from "installation lookup failed" (raise / surface) in `_from_installation`; feed the failure into a health signal.
3. Tighten the `AGENTS.md` score-formula wording to match `analytics_service.py`.

nazarli-shabnam:
## What

Fixes #414.

Two catch-all exception handlers were turning real outages into silent
behavior changes, plus one doc gap:

1. `app_config.get_config` caught every exception and reverted to the code
   default. A transient DB blip, or a missing `SELECT` grant under a
   constrained `clevis_api`/`clevis_worker` role, would silently flip
   `registration_enabled` back to `"true"` (re-opening self-registration) and
   reset every poll interval / `pr_nudge_mode` / `digest_cadence` to their
   code defaults -- with only a `warning` log line.
2. `token_resolution._from_installation` logged a genuine installation-token
   mint failure (rotated/revoked `GITHUB_APP_PRIVATE_KEY`, GitHub 5xx) at the
   same `warning` level as the routine "App isn't configured yet" case,
   silently degrading every request through that installation to a
   client-supplied-PAT fallback with no way to tell the two apart from logs.
3. `AGENTS.md`'s score-formula description didn't match what
   `analytics_service.py` actually computes (doesn't account for `"error"`
   counting as failed, or `"not_applicable"` being excluded from the
   denominator).

## Fix

1. `get_config` now serves the last-known-good cached value on a read
   failure (falling to the code default only if nothing has ever been read
   successfully for that key), and logs at `error`. Doesn't refresh the stale
   cache entry's timestamp, so the next call retries the DB rather than
   pinning the stale value for a full TTL.
2. `_from_installation`'s mint-failure branch now logs at `error` instead of
   `warning` -- distinct from the `GitHubAppNotConfigured` branch, which
   stays at `warning` since that's a routine, expected state until an
   operator sets the App up. No change to the actual fallback/control flow.
3. Tightened the `AGENTS.md` wording to match `analytics_service.py`'s real
   formula.

**Deliberately not done:** issue #414's suggestion to feed these into a
health signal / `/healthz`. `/healthz` is currently a static
`{"status": "ok"}` with zero signal plumbing -- building that out is a
separate feature, not part of this silent-failure fix. Filing a follow-up
issue if this isn't already tracked would make sense, but is out of scope
here.

## Testing

- Added `apps/api/tests/test_app_config.py`: covers the success path, the
  no-cache/fall-to-default path, the has-cache/serve-stale-value path
  (the actual regression), and the still-fresh-cache/no-DB-call path.
- Added `test_installation_mint_failure_logs_at_error_not_warning` to
  `apps/api/tests/test_token_resolution.py`, asserting the log level via
  `caplog`.
- Full suite: `pytest -q` -- 1198 passed, 3 skipped.
- `bun run check` / `bun run test` clean via the pre-push hook (this PR
  doesn't touch UI code, but the hook runs regardless).

No migrations, no RBAC change. Touches credential-resolution logic
(`token_resolution.py`) -- flagging per AGENTS.md's "sensitive files" note,
though the change is purely to log severity, not to the resolution/fallback
logic itself.


nazarli-shabnam:
### Code review

Found 1 issue (fixed in a follow-up commit before you look at this):

1. Stale log-level reference in `_no_token_error`'s comment -- it said "see the warning `_from_installation` already logged", but this PR's own change bumped that log call to `logger.error`.

https://github.com/nazarli-shabnam/clevis/blob/bf76a0b3fb6889987703832c79c326afbf26e370/apps/api/src/services/token_resolution.py#L69-L74

Fixed in 19fc1c1.

nazarli-shabnam:
@coderabbitai review pls detailed

PR #443, opened 2026-09-16
feat(db): force row-level security on the remaining S3-S6 tenant tables

### LINKED ISSUE #412: feat(db): finish RLS FORCE rollout + enforce tenant isolation in the default deployment
_Found by an audit sweep on `main` @ `31ff8ce` (2026-09-09). Broader than the now-closed #330, which was only about the API's role identity._

### What
Row-Level Security was scaffolded across S2–S6 but the "enable now, `FORCE` later" plan has stalled, and **in a default deployment RLS enforces nothing**.

- Migrations `0030`, `0036`, `0037`, `0038`, `0039`, `0040`, `0041`, `0043` all `ENABLE ROW LEVEL SECURITY` **without `FORCE`**. Only `0031` adds `FORCE` — to just `memberships`, `github_installations`, `scan_results`.
- `docker/postgres-init/01-create-worker-role.sh` / `02-create-api-role.sh` run only on a fresh data volume and only if `WORKER_DB_PASSWORD` / `API_DB_PASSWORD` are set. A default deploy sets neither, so both API and worker connect as the `initdb` bootstrap **superuser**, which bypasses RLS unconditionally regardless of `ENABLE`/`FORCE`.

So tenant isolation for `repo_events`, `repo_event_daily_counts`, `security_alerts`, `org_members`, `repo_collaborators`, `activity_sync_cursors`, `org_membership_sync_cursors`, `automation_repo_settings`, … rests **entirely** on application-level `WHERE tenant_id = :t`. RLS is defense-in-depth that is off by default and, for the non-`FORCE`d tables, also off for the table owner even after the `clevis_api` cutover.

### Also
The only end-to-end "RLS blocks cross-tenant reads" tests (`apps/api/tests/test_rls_isolation.py`, `test_installation_deleted_rls_fix.py`) `pytest.skip` unless the run points at `clevis_api`. A default local `pytest -q` (the documented path) never exercises enforcement — regression coverage for "isolation works" is a single CI job.

### Suggested fix
1. A deliberate, reviewed migration adding `FORCE ROW LEVEL SECURITY` to the remaining S3–S6 tenant tables.
2. Decide + document whether a default deployment should auto-provision `clevis_api` / `clevis_worker` so RLS is actually on out of the box. If it stays opt-in, state plainly in README / self-hosting that tenant isolation is app-layer-only otherwise.
3. Consider making the RLS-isolation tests run by default (provision the non-superuser role in the fixture) instead of skipping.

nazarli-shabnam:
## Schema change included -- read before merging

**This PR includes migration `0046`, which adds `FORCE ROW LEVEL SECURITY` to 8
existing tables.** No column/type changes, no data movement -- purely an RLS
enforcement-mode flag. Reversible (`downgrade` sets `NO FORCE`).

## What

Fixes #412 (items 1 and 2 of its suggested fix; item 3 addressed separately below).

Migrations `0036`, `0037`, `0038`, `0039`, `0040`, `0041`, `0043` each `ENABLE ROW
LEVEL SECURITY` on their table but never add `FORCE` -- the same gap migration `0031`
already described and fixed for `memberships`/`github_installations`/`scan_results`.
Without `FORCE`, RLS policies don't apply to a table's *owner* (the API's own DB role
in every deployment, since Alembic runs as it), so these 8 tables' RLS was inert for
that connection the entire time -- "enable now, `FORCE` later" stalled after `0031`
and never got to the rest.

## Why this is safe

Before adding `FORCE`, I audited every write path to these 8 tables the same way
`0031`'s own audit did for `memberships`/`github_installations` (its docstring is
explicit about *why* that audit mattered: `FORCE`'s implicit `WITH CHECK` would reject
any write that doesn't already satisfy the tenant-equality policy):

- `repo_events`, `repo_event_daily_counts`, `activity_sync_cursors`, `security_alerts`
  -- written from `apps/worker` after an explicit `SET app.tenant_id = <n>` on the same
  cursor, matching the row being written (verified in `worker.py`'s backfill handler
  and `event_consumer.py`'s security-alert branch).
- `org_members`, `repo_collaborators`, `org_membership_sync_cursors` -- same: the
  caller sets `app.tenant_id` on the cursor immediately before calling into
  `org_membership_store.py` (verified in both `event_consumer.py`'s member/organization
  branch and `worker.py`'s membership-reconcile handler).
- `automation_repo_settings` -- written only from `require_org_role`-gated request
  handlers, which already call `set_tenant_session_context` before the handler body
  runs.

No self-access policy widening needed (unlike `0031`) -- every write already goes
through explicit tenant-context setting, and all eight tables have a plain
`tenant_id`-`NOT NULL` equality policy (no `OR NULL` carve-out to preserve).

**Verification:** ran the migration locally, then manually provisioned a throwaway
non-owner role (same shape as `clevis_api`) and confirmed the existing RLS isolation
tests (`test_rls_isolation.py`, `test_installation_deleted_rls_fix.py`) still pass
under it -- full suite: 1194 passed, 3 skipped -- then dropped the role. `FORCE`'s
marginal effect on the actual table *owner* can't be observed in this local
docker-compose setup, since `DB_USER` there is a Postgres superuser and superusers
bypass RLS unconditionally regardless of `ENABLE`/`FORCE` -- same limitation `0031`'s
own docstring already accepted and disclosed (it verified against "a manually-created
non-superuser test role, not the app's real connection").

## The architectural decision (item 2 of the issue)

Deliberately did **not** change any default. `API_DB_PASSWORD`/`WORKER_DB_PASSWORD`
still default to unset, in which case RLS (`FORCE` or not) is fully inert and tenant
isolation is app-layer-only, same as before this PR. Making the non-superuser roles
required by default would be a breaking change to every existing deployment (new
required env vars, or the app fails to start / silently misbehaves for anyone
upgrading in place) -- explicitly out of scope here per discussion before starting
this PR.

Instead, documented the gap plainly per the issue's own fallback option:
- `docs/self-hosting.md`'s "Security notes" section now states outright that a default
  deployment's tenant isolation is app-layer-only.
- Step 6 (`API_DB_PASSWORD`) now explains that `FORCE` alone changes nothing without it.
- `AGENTS.md`'s `API_DB_PASSWORD` entry now mentions migration `0046` and links to the
  same caveat.

## Item 3 (RLS tests skip by default)

Not changed in this PR -- making `test_rls_isolation.py` provision its own
non-superuser role automatically (instead of skipping without `API_DB_PASSWORD`) is
itself the same category of "change a default" decision as item 2, and the existing
skip is already documented as deliberate ("the documented, optional 'Recommended'
setup per AGENTS.md, not a hard requirement"). Left it consistent with the "keep
opt-in" choice above rather than deciding unilaterally to change test infra behavior.

Added a different, always-on regression check instead:
`test_every_tenant_table_has_force_rls` queries `pg_class.relforcerowsecurity`
directly (no special role needed, never skips) and asserts every table that should
have `FORCE` does -- covering both `0031`'s three tables and this PR's eight.

## Testing

- `alembic upgrade head` applied cleanly.
- Full suite: `pytest -q` -- 1194 passed, 3 skipped.
- Manual non-superuser-role verification described above.

Does not touch RBAC code, `.env`, or secrets. Migration is additive/reversible, no
backfill risk (no data touched, only an enforcement-mode flag).

## Summary by CodeRabbit

* **Security**
  * Tenant-scoped database tables now enforce row-level security for configured non-superuser API and worker roles, strengthening tenant isolation.

* **Documentation**
  * Clarified that row-level security remains opt-in and that default deployments rely on application-layer tenant filtering unless the required database credentials are configured.


nazarli-shabnam:
### Code review

Found 1 issue, addressed below (no code change needed -- this was a verification gap, not a bug):

1. Migration 0031's precedent (and migration 0033's own history) is to actually exercise writes under a real non-owner role with the policy active, not just audit the code by reading it -- 0033's docstring notes its bugs (broken `USING` clauses, rejected legitimate writes) were "invisible until this cutover" to a real role. My earlier verification only ran the *existing* RLS isolation tests (`memberships`/`github_installations`/`audit_logs`), not a functional write test against any of the 8 *new* tables this migration adds `FORCE` to.

Closed the gap by hand: created a throwaway non-owner role, and ran the actual write patterns from the real call sites against three representative tables under `FORCE`:

- `repo_events` (worker's `SET app.tenant_id` → `INSERT` pattern): a legitimate insert (`tenant_id=1` under `app.tenant_id=1`) succeeded; an insert with a mismatched `tenant_id=2` under the same session was rejected with `new row violates row-level security policy for table "repo_events"`.
- `automation_repo_settings` (API's `require_org_role` → upsert pattern): same result for a plain insert, and specifically for the `ON CONFLICT (...) DO UPDATE` shape `automation_settings_repo.upsert` actually uses.
- `org_members`: seeded one row each for two tenants, then ran a **broad** `DELETE FROM org_members WHERE login NOT IN (...)` under tenant 1's session context -- a query that would delete both tenants' rows without RLS. Only tenant 1's row was deleted; tenant 2's row was confirmed intact both from a tenant-2-scoped session and as the superuser (bypassing RLS for a true global view). This is the scenario 0033 showed static auditing alone can miss.

All test data and the throwaway role were removed afterward; `pytest -q` confirms a clean state (1194 passed, 3 skipped).

PR #445, opened 2026-09-16
chore(ci): actually run the built images instead of importing them

### LINKED ISSUE #418: chore(ci): CI never runs the built release images; UI image has no smoke test
_Found by an audit sweep on `main` @ `31ff8ce` (2026-09-09). Matches the standing "Docker Build Verification" follow-up note; no existing issue._

### What
`.github/workflows/ci.yml`'s `docker` job **builds** the API, Worker, and UI images (`docker build`) and then, for API and Worker only, runs a bare `docker run --entrypoint python … -c "import src.main"` / `import worker`. It never:
- executes `entrypoint.sh` (env-var validation, URL construction, `alembic upgrade head`, the `clevis_api` role switch),
- boots `uvicorn` / the worker poll loop against a DB,
- smoke-tests the **UI image at all** (no run, no `curl`).

The only place a real running stack exists is the `e2e` job — and that stack is built fresh from source via `docker compose --build`, **not** from the images `release.yml` actually ships. So a regression in `entrypoint.sh`, the Dockerfile runtime stage, the standalone Next build, or image-only dependency resolution passes CI and is first discovered on deploy.

### Suggested fix
Add a step to the `docker` job (or a new job) that runs each built image the way it runs in production: `docker run` the API image against a throwaway Postgres/Redis and `curl /healthz`; `docker run` the worker image and assert it reaches "worker started"; `docker run` the UI image and `curl /login`. Or point the `e2e` job at the tagged images from the `docker` job instead of `--build`.

nazarli-shabnam:
## What

Fixes #418.

`.github/workflows/ci.yml`'s `docker` job built the API/Worker/UI images, then only ran
`docker run --entrypoint python -c "import src.main"` / `"import worker"` against dummy
env vars with no live DB. It never executed `entrypoint.sh` (env-var validation,
`alembic upgrade head`, the `clevis_api` role switch), booted `uvicorn` / the worker poll
loop against a real DB, or smoke-tested the UI image at all. The only place a real
running stack existed was the `e2e` job -- built fresh from source via
`docker compose --build`, not from the images this job actually verified.

## Fix

Replaced the build+import checks with `docker compose ... up --build -d --wait`, which
blocks until every service's own healthcheck (already defined in `docker-compose.yml` --
API's calls its own `/healthz`, worker's checks its heartbeat files, UI's curls
`/api/health`) reports healthy, or fails the job if one doesn't. This runs each image
exactly the way it runs in production, against a real throwaway Postgres/Redis from
`docker-compose.ci.yml` -- not just proving the Dockerfile's instructions succeed.

Updated `CONTRIBUTING.md`'s "Docker images" section to match, with the equivalent local
command.

## Testing (and a real bug this approach caught)

Ran the exact CI command locally (`docker compose -f docker-compose.yml -f
docker-compose.ci.yml up --build -d --wait`) before opening this PR. It failed --
`alembic` refused to start because my local dev DB had a leftover `alembic_version` row
from testing an unrelated PR branch that isn't on this one. That's exactly the class of
runtime failure the old build-only check could never have caught (a mismatched migration
state, or any other "the image builds but doesn't actually come up" bug). Fixed locally
(not part of this PR's diff), then reran and confirmed all 5 services reach `healthy` and
`/healthz` / `/login` both return `200`.

- `docker compose -f docker-compose.yml -f docker-compose.ci.yml up --build -d --wait` --
  all 5 services healthy.
- Full suite: `pytest -q` -- 1195 passed, 3 skipped.

No migrations, no RBAC change, no code behavior change -- CI/docs only.

## Summary by CodeRabbit

- **CI/CD**
  - Updated Docker verification to build and start the complete application stack together.
  - CI now waits for API, worker, and UI health checks to pass before succeeding.
  - Added automatic service log collection when verification fails.
  - Ensured containers and volumes are always cleaned up after the job.

- **Documentation**
  - Updated contributor instructions with the new Docker Compose verification workflow and local commands.


nazarli-shabnam:
### Code review

Found 2 issues:

1. Inaccurate "curls" wording -- the API healthcheck runs `python -c "import urllib.request; ..."` and UI's runs `node -e "require('http').get(...)"`; neither calls `curl` (`docker-compose.yml`'s own comment on the UI healthcheck already explains why: no curl/wget in the `node:22-alpine` runner). Fixed in a follow-up commit before you look at this.

https://github.com/nazarli-shabnam/clevis/blob/35c65b90a69c5f69ec59f5d4abe86d942fd9851b/.github/workflows/ci.yml#L316-L319

2. Not fixed -- a real tradeoff worth flagging, not a bug: the original "docker" job was deliberately kept lightweight (build + bare import, no live DB) specifically so it stayed fast, with the `e2e` job doing the one heavy full-`--build`-plus-Playwright run. This PR makes `docker` do the same full `docker compose ... --build` as `e2e`, just without Playwright -- so every PR now does two independent full image builds instead of one. The issue's own suggested fix offered an alternative that avoids this ("point the e2e job at the tagged images from the docker job instead of --build"), which would need image-sharing between jobs (registry push/pull or artifacts) -- a meaningfully bigger change than this PR's scope. Accepting the extra CI-minutes cost as the tradeoff for actually testing runtime behavior seems reasonable, but it's worth being aware of, and revisiting if CI cost becomes a real concern.

https://github.com/nazarli-shabnam/clevis/blob/35c65b90a69c5f69ec59f5d4abe86d942fd9851b/.github/workflows/ci.yml#L211-L237

Issue #282, opened 2026-07-25
/collaborators is a redirect stub — merge into or replace with settings org members page


nazarli-shabnam:
`apps/ui/app/collaborators/page.tsx:24-26` has no real content of its own — it immediately `router.replace`s to `/settings/org/${org}/members` once it finds an admin org (or shows an empty/scolding state if you're not an org admin). All actual functionality (members table, 2FA status, outside collaborators, permission audit, inactive members, invitations) lives entirely in `apps/ui/app/settings/org/[login]/members/page.tsx` (`GithubRoster`, lines 24-401).

A top-level sidebar nav item that exists solely to bounce into a nested Settings route is confusing IA.

**Recommended direction:** either (a) point the sidebar nav item directly at `/settings/org/[login]/members` and remove the intermediate `/collaborators` page, or (b) hoist that Settings page's content up to become the real `/collaborators` page and redirect the old Settings sub-route instead.

Design/build item for a future session — not scoped for immediate implementation.

PR #449, opened 2026-09-17
fix(ui): /automation renders with the wrong document title

### LINKED ISSUE #421: fix(ui): /collaborators and /automation render with the wrong document <title>
_Found by an audit sweep on `main` @ `31ff8ce` (2026-09-09), live click-through of every nav page. Minor UX / a11y._

### What
`/collaborators` and `/automation` render with the browser tab / document `<title>` **"Overview · clevis"** instead of their own page name. Every other top-level page has an `app/<page>/layout.tsx` with `export const metadata = { title: "… · clevis" }` (`security`, `repos`, `activity`, `releases`, `pulls`, `audit`, `my`) — `collaborators/` and `automation/` have no `layout.tsx` and no `metadata` export, so they inherit the root layout's `metadata.title` (`apps/ui/app/layout.tsx:32-33`).

The roadmap's own Cross-Cutting Concerns section lists "`<title>` per page" as a requirement, so this is a known standard the two pages miss.

Related: the root `metadata.title` is a literal `"Overview · clevis"` rather than a `title.template` + neutral `default` — which is why the fallback reads as the Overview page rather than a generic "clevis". Worth switching to `{ default: "clevis", template: "%s · clevis" }` so any future page without its own title degrades sanely.

### Repro
Log in, visit `http://localhost:3000/collaborators` or `/automation`, check the tab title.

### Suggested fix
Add `app/collaborators/layout.tsx` and `app/automation/layout.tsx` with the appropriate `metadata.title`; optionally convert the root `metadata` to `title.template` form.

nazarli-shabnam:
## What
`/automation` had no `app/automation/layout.tsx`, so it inherited the root layout's `metadata.title` ("Overview · clevis") instead of its own page name in the browser tab.

## Scope note
Issue #421 also named `/collaborators`, but that route no longer exists — #282 (closed) merged it into `/settings/org/[login]/members`, which already has its own title via `app/settings/layout.tsx`. So this PR only needed the `/automation` half.

The issue also suggested optionally converting the root layout's `metadata.title` to `{ default, template }` form. Skipped: every other page's layout already bakes the full `"<Page> · clevis"` string into its own `title`, so switching the root to a template would double-suffix all 12 of them unless each one's title string were also rewritten — a much bigger diff than this bug needs. Left as a separate, larger cleanup if wanted.

## Change
Added `apps/ui/app/automation/layout.tsx` with `metadata.title = "Automation · clevis"`, matching the existing pattern (`app/security/layout.tsx`, `app/audit/layout.tsx`, etc.).

Closes #421

## Summary by CodeRabbit

* **New Features**
  * Added an Automation page layout.
  * Set the Automation page title to “Automation · clevis.”


nazarli-shabnam:
### Code review

No issues found. The new `apps/ui/app/automation/layout.tsx` matches the existing per-page title pattern exactly (e.g. `app/repos/layout.tsx`). No `dynamic = "force-dynamic"` export is needed here — `/automation`'s page is client-rendered (`"use client"`, data fetched via `useQuery`) and was already statically prerendered (`○`) in the build output both before and after this change, same as `/repos` which uses the identical layout shape without that export.

nazarli-shabnam:
### Code review

No issues found. Verified the fix matches sibling patterns and fully covers what issue #421 describes.

PR #451, opened 2026-09-17
chore(deps): align cryptography and pydantic-settings between api and worker

### LINKED ISSUE #419: chore(deps): align cryptography / pydantic-settings between api and worker
_Found by an audit sweep on `main` @ `31ff8ce` (2026-09-09). Low severity, latent risk._

### What
The API and worker pin different versions of two shared-surface libraries:

| package | `apps/api/requirements.txt` | `apps/worker/requirements.txt` |
|---|---|---|
| `cryptography` | `44.0.2` | `48.0.1` |
| `pydantic-settings` | `2.14.1` | `2.14.2` |

(`pydantic` itself matches at `2.11.9`.)

The API and worker share Fernet token crypto through `checks.crypto` (`AGENTS.md` explicitly calls the api/worker `_crypto.py` wrappers a "keep in sync" sharp edge). A four-major-version gap in `cryptography` between the two processes that encrypt/decrypt the same job-payload tokens is exactly the kind of drift that "keep in sync" is meant to prevent — today it happens to be compatible, but nothing enforces that.

Neither requirements file pins `clevis-checks` (installed via `pip install -e packages/checks` / `pip install /app/packages/checks` in the Dockerfiles), so the shared crypto lib's own transitive deps aren't locked either.

### Suggested fix
Align `cryptography` and `pydantic-settings` across `apps/api/requirements.txt` and `apps/worker/requirements.txt` (bump the API to match the worker, or vice-versa, after a test run). Consider a shared constraints file, or moving the common runtime deps into `packages/checks`'s own metadata so both services inherit one pinned set.

nazarli-shabnam:
## What
`apps/api/requirements.txt` pinned `cryptography==44.0.2` and `pydantic-settings==2.14.1`, while `apps/worker/requirements.txt` pinned `cryptography==48.0.1` and `pydantic-settings==2.14.2`.

## Why
The API and worker share Fernet token crypto through `checks.crypto` (`AGENTS.md` explicitly calls the api/worker `_crypto.py` wrappers a "keep in sync" sharp edge) — a four-major-version gap in the library that encrypts/decrypts the same job-payload tokens between the two processes is exactly the kind of drift that note is meant to prevent. It happened to still be compatible, but nothing enforced that.

Also found while verifying: `packages/checks/pyproject.toml` already pins `cryptography==48.0.1` directly, so the worker's version was already the one actually shared by the crypto lib both services install alongside — the API's `44.0.2` was the outlier, not a deliberate choice.

## Change
Bumped `apps/api/requirements.txt`'s `cryptography` to `48.0.1` and `pydantic-settings` to `2.14.2`, matching both the worker and `packages/checks`'s own pin. Verified locally: `pip install -r apps/api/requirements.txt` (with `packages/checks` reinstalled editable first) resolves with no conflicts, and `pytest apps/api/tests/test_crypto.py` (Fernet encrypt/decrypt round-trip) passes unchanged against the new `cryptography` version.

Not done: the issue also suggested pinning `clevis-checks` itself in each requirements file or moving shared runtime deps into `packages/checks`'s own metadata. Skipped — `clevis-checks` is installed editable from a local path (`pip install -e packages/checks`), not a versioned PyPI release, so there's no meaningful version string to pin there; that's a bigger packaging change than this issue's core ask (the two `cryptography`/`pydantic-settings` versions actually drifting) needs.

Closes #419

## Summary by CodeRabbit

* **Chores**
  * Updated underlying security and configuration components to newer versions for improved compatibility and maintenance.


nazarli-shabnam:
### Code review

No issues found. Verified the version bump matches both apps/worker/requirements.txt and packages/checks/pyproject.toml's existing pin, and the diff is minimal.

PR #453, opened 2026-09-17
fix(worker): serialize membership reconciliation against concurrent webhook mutations

### LINKED ISSUE #357: Prevent membership reconciliation from clobbering concurrent webhook mutations
## Summary

Prevent organization-membership reconciliation from overwriting webhook mutations that occur while the worker fetches a GitHub organization roster.

## Rationale

`apps/worker/src/worker.py` fetches a complete roster through one or more GitHub API calls and then applies an authoritative snapshot in `apps/worker/src/org_membership_store.py`. A webhook can add or remove a member after the fetch starts and before the snapshot write completes.

This race can recreate a removed member or delete a newly added member. Timestamp comparison on an existing row is insufficient because a removal can delete that row before the snapshot is applied.

## Affected areas

- `apps/worker/src/worker.py`: `_handle_reconcile_org_membership`
- `apps/worker/src/membership_reconcile.py`: roster fetch flow
- `apps/worker/src/org_membership_store.py`: webhook mutations and `reconcile_org_members`
- Relevant worker and store tests

## Required changes

Implement one design that preserves webhook ordering across the full roster-fetch and snapshot-apply interval:

- Use a shared per-tenant mechanism that serializes membership webhook mutations with reconciliation from before the roster fetch until after the snapshot transaction commits; or
- Use durable snapshot-watermark and tombstone/version semantics that cause the snapshot apply to preserve webhook mutations newer than its watermark, including removals.

Do not rely only on `added_at`, `granted_at`, or comparisons against currently existing membership rows.

## Acceptance criteria

- A member removal webhook that arrives after the roster fetch begins cannot be recreated by the reconciliation snapshot.
- A member addition webhook that arrives after the roster fetch begins cannot be deleted by the reconciliation snapshot.
- Removal state remains durable enough to compare against a later snapshot.
- The mechanism covers the entire roster-fetch and snapshot-apply interval, not only the database write.
- The worker handles failures without leaving the tenant permanently blocked or advancing a synchronization cursor for an unapplied reconciliation.
- Tests demonstrate both concurrent mutation cases with deterministic synchronization and verify normal reconciliation still completes.

## Backlinks

- Source PR: https://github.com/nazarli-shabnam/clevis/pull/356
- Source review discussion: https://github.com/nazarli-shabnam/clevis/pull/356#discussion_r3871729397

Requested by: @nazarli-shabnam

nazarli-shabnam:
## What
The org-membership reconciliation job (\`worker.py\`'s \`_handle_reconcile_org_membership\`) fetches a full org roster from GitHub, then applies it as an authoritative snapshot in \`org_membership_store.reconcile_org_members\` — including a \`DELETE ... WHERE NOT (login = ANY(seen_logins))\` for anyone the roster didn't include. A \`member_added\`/\`member_removed\` webhook landing between the roster fetch starting and the snapshot commit could be silently undone: a member added mid-fetch isn't in the (already-fetched) roster, so the snapshot's DELETE removes them right back out; a member removed mid-fetch is still in the (stale) roster, so the snapshot's INSERT resurrects them. The existing per-row \`added_at\`/\`granted_at\` ordering guard in the webhook write path can't catch this, because the snapshot's DELETE has no per-row timestamp to compare against at all.

## Design decision
Issue #357 laid out two options: (a) a shared per-tenant mechanism serializing webhook mutations against reconciliation for the whole fetch-to-commit window, or (b) durable snapshot-watermark + tombstone semantics. I went with **(a)**: a session-level Postgres advisory lock, keyed \`(hashtext('org_membership_reconcile'), tenant_id)\` — same two-key convention \`apps/api/src/services/sweep_lock.py\`'s \`try_acquire_sweep_slot\` already uses for a different lock family, but blocking (\`pg_advisory_lock\`, not the \`_try\` variant) and session-scoped (not \`_xact\`) since it has to span an HTTP round trip to GitHub that happens before any transaction opens.

(b) would need a new tombstone table and version/watermark columns — a real schema change for a low-frequency event (org membership changes aren't a hot path). (a) reuses an existing pattern in this codebase, needs no migration, and directly satisfies the issue's acceptance criteria (serializes the *entire* fetch-to-commit window, not just the DB write).

## Change
- \`org_membership_store.py\`: new \`acquire_tenant_lock\`/\`release_tenant_lock\` helpers.
- \`worker.py\`'s \`_handle_reconcile_org_membership\`: acquires the lock before \`fetch_org_roster\`'s GitHub call, releases in a \`finally\` covering every return path (rate-limit requeue, permanent failure, incomplete-roster requeue, DB-error rollback, and success).
- \`event_consumer.py\`'s \`member\`/\`organization\` webhook-event branch: acquires the same per-tenant lock before its write, releases in a \`finally\`.
- Fixed 8 existing \`_FakeConn\`-based tests in \`test_membership_reconcile.py\` that asserted on \`calls[0]\` — they now mock out the lock helpers so the assertion still targets the actual status-update SQL, not the lock's own statement.
- New \`test_org_membership_store.py\`: verifies the lock actually blocks a second real connection until the first releases, and that it's reentrant on the same connection (the reconcile handler holds it across multiple statements on one connection).

## Trade-off, stated explicitly
A webhook event for a tenant currently being reconciled now blocks behind that reconciliation (bounded by however long the GitHub roster fetch + snapshot write takes — typically seconds). Since \`event_consumer.py\`'s stream consumer is single-threaded, this can briefly delay *other* tenants' pending membership-webhook entries behind it too. Accepted as the direct cost of "serializes... until after the snapshot transaction commits," which the issue explicitly asked for; membership webhooks are low-frequency enough that this shouldn't be a practical bottleneck.

## Verification
\`pytest apps/worker/tests\` — 169 passed, including the new lock tests and the full existing \`test_membership_reconcile.py\`/\`test_event_consumer.py\` suites. Full repo \`pytest -q\` — 1217 passed, 3 skipped.

Closes #357

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved organization and membership updates to prevent concurrent operations from conflicting for the same tenant.
  * Ensured processing locks are released after database errors, allowing affected events to be retried safely.
  * Improved transaction handling so failed updates do not leave processing blocked.

* **Tests**
  * Added coverage for lock coordination, reentrant locking, and lock release after database failures.


nazarli-shabnam:
### Code review

Found 1 issue:

1. **`release_tenant_lock` in the `finally` can itself fail and leak the advisory lock when the guarded write raises `psycopg.Error`.** In `event_consumer.py`'s org-membership webhook branch, the DB writes inside `with pg_conn.cursor() as cur:` (https://github.com/nazarli-shabnam/clevis/blob/6f3ba2d47b17e8dde2c78c1912e3e896d3ece776/apps/worker/src/event_consumer.py#L369-L405) have no local `except psycopg.Error` — unlike `worker.py`'s reconcile handler, which explicitly does `conn.rollback()` before returning on a DB error. If any statement here raises (e.g. a deadlock or serialization failure — plausible precisely because this code path now runs concurrently with the reconcile job under lock contention), the connection enters Postgres's aborted-transaction state. The `finally` block then calls `release_tenant_lock` (https://github.com/nazarli-shabnam/clevis/blob/6f3ba2d47b17e8dde2c78c1912e3e896d3ece776/apps/worker/src/event_consumer.py#L408-L409), whose `cur.execute(\"SELECT pg_advisory_unlock(...)\")` (https://github.com/nazarli-shabnam/clevis/blob/6f3ba2d47b17e8dde2c78c1912e3e896d3ece776/apps/worker/src/org_membership_store.py#L58-L63) is an ordinary statement, not COMMIT/ROLLBACK — Postgres rejects it with `InFailedSqlTransaction` while the transaction is aborted. That new exception replaces the original one, `conn.commit()` is never reached, and the unlock never happens. The caller's outer `except Exception: _rollback_quietly(pg_conn)` (event_consumer.py's `run()`/`_sweep_pending`) does clear the SQL transaction state afterward, but a plain `ROLLBACK` does not release a session-level `pg_advisory_lock` — only an explicit unlock or closing the connection does. Net effect: the per-tenant lock stays held for the rest of that connection's lifetime (the remainder of the current batch), so a concurrent reconcile job or another consumer for the same tenant blocks until the batch's connection closes, undermining the exact serialization guarantee this PR adds. `release_tenant_lock`'s docstring (\"Safe to call after the same connection's pending transaction was rolled back\") is true only if a rollback already happened *before* it runs — it doesn't hold when `release_tenant_lock` itself is the first thing to touch the aborted connection, which is exactly this `finally` path. Fix: catch `psycopg.Error` around the writes in this branch (or in `release_tenant_lock` itself) and roll back before attempting the unlock, mirroring `worker.py`'s pattern.

Everything else checked out: every return path in both `worker.py`'s `_handle_reconcile_org_membership` and `event_consumer.py`'s org-membership branch does route through the lock's `finally`/release on the happy path; the lock's dual-key namespace (`hashtext('org_membership_reconcile')`, tenant_id) doesn't collide with `sweep_lock.py`'s (different job_type strings feeding `hashtext`) or with `auth.py`/`installations.py`'s single-key `pg_advisory_xact_lock` calls (a separate key space in Postgres); and the new `test_org_membership_store.py` threading test genuinely proves cross-connection serialization via a real blocked-then-released DB lock, not a timing guess.

nazarli-shabnam:
Fixed in 9a914d6: the org-membership webhook branch in `event_consumer.py` now catches `psycopg.Error`, rolls back, and re-raises before the `finally`'s `release_tenant_lock` runs — matching `worker.py`'s existing pattern. Added `test_organization_event_db_error_still_releases_the_tenant_lock`, which proves a second connection can immediately re-acquire the tenant's lock after a simulated write failure. Full suite (1218 tests) passes.

nazarli-shabnam (commented):
@coderabbitai None of the referenced patterns (_raise monkeypatch, UniqueViolation, DivisionByZero, release_tenant_lock, acquire_tenant_lock) exist anywhere in this codebase. Let me double check the reconcile test file too and confirm there's no tenant-lock concept at all in this repo.

  Ran 1 shell command

All three findings reference code that doesn't exist in this repo — verified against the actual files, not just descriptions:

- worker.py:485-486 (process_job/psycopg.Error handler): Line 485-486 is actually the JOB_HANDLERS dict, not process_job. process_job (line 170) already has a blanket except Exception that calls _mark_failed, and neither process_job nor anything else in the file has a finally block or a release_tenant_lock/tenant-advisory-lock concept — grep -rn "tenant_lock" apps/worker apps/api returns nothing. This finding describes a function/mechanism that isn't in the codebase. Skipped.
- test_membership_reconcile.py:293-414 (lock-interleaving regression tests): Same issue — there is no per-tenant advisory lock shared between _handle_reconcile_org_membership and event_consumer._process_entry to interleave against. Tenant isolation there is via SET app.tenant_id session var + RLS, not a lock. The actual lines 280-300 are unrelated failure-path tests (test_handler_marks_failed_on_undecryptable_token, etc.). Skipped — no such lock exists to test.
- test_event_consumer.py:866-867 (_raise monkeypatch → DivisionByZero): Those lines are part of test_process_entry_drops_malformed_json_payload, inserting a delivery row — no _raise monkeypatch, UniqueViolation, or DivisionByZero anywhere in the file (grep confirms). Skipped.

No code changes made — nothing in these three comments matches current code, and I'm not going to invent a lock mechanism or exception path to satisfy a stale/fabricated review. If you have a specific PR/commit these comments were meant for, point me at it and I'll re-check against that diff.

nazarli-shabnam:
@coderabbitai see the comment above?

Issue #409, opened 2026-09-09
fix(api): webhook_deliveries rows stuck at status='queue_failed' are never re-enqueued


nazarli-shabnam:
_Found by an audit sweep on `main` @ `31ff8ce` (2026-09-09). Not-yet-built path, no existing tracking issue._

### What
When a verified GitHub webhook arrives but the Redis `XADD` throws (Redis unreachable / auth failure / OOM), `apps/api/src/routers/webhooks.py` (~L195-201) still stores the delivery but sets `webhook_deliveries.status = 'queue_failed'` instead of `'queued'`. **Nothing ever reads that status back.** `grep -rn queue_failed apps/` finds only the write site and the model comment.

`apps/api/src/core/db.py:146` documents the intent — "lets a future re-enqueue sweep find rows" — and `ix_webhook_deliveries_status` exists specifically to make that sweep cheap. The sweep was never built.

### Impact
If Redis is down for any window, every webhook received in that window is durably stored but permanently absent from the activity feed / security-alert / collaborator pipelines. Silent — no alert, no retry, no operator signal. The point of persisting before enqueue (S3 design) is defeated at the last step.

### Suggested fix
A periodic sweep (in the worker alongside `_reclaim_stale_jobs`, or an API asyncio loop like `gap_heal_loop`) that re-`XADD`s `status='queue_failed'` rows and flips them to `'queued'` on success, with a bounded retry / age cap and a log line when the backlog is non-empty. Add a test — there is none today, because the feature is absent.

PR #454, opened 2026-09-17
chore(worker): bound the webhook_events Redis stream with MAXLEN trimming

### LINKED ISSUE #440: chore(worker): webhook_events Redis stream has no retention/trimming policy
## What

The \`webhook_events\` Redis Stream (written by \`apps/api/src/routers/webhooks.py\`'s \`_handle_ingested_event\` and, as of #409, also by \`apps/api/src/services/webhook_requeue_sweep.py\`) has no defined retention or trimming policy. Entries accumulate forever unless something reads and ACKs them off the consumer group's pending-entries-list (\`apps/worker/src/event_consumer.py\`'s \`XCLAIM\`/\`XACK\` handle acking, but the stream itself is never \`XTRIM\`med).

Originally flagged by CodeRabbit on PR #338 (which introduced the stream / \`queue_failed\` status): "defer the XADD call until the \`webhook_events\` stream has an established consumer, recovery, and retention lifecycle... do not introduce trimming without a defined recovery policy for \`delivery_row_id\` references." That concern was never addressed, and #409's requeue sweep adds a second producer onto the same untrimmed stream.

## Why this matters

Unbounded Redis memory growth over the life of a deployment. Not urgent (no incident reported), but worth a deliberate decision before it becomes one -- especially now that two independent code paths write onto the stream.

## Suggested approach

Decide a retention policy (time-based \`MAXLEN\` approximate trim via \`XADD ... MAXLEN ~ N\`, or a periodic \`XTRIM\`/\`XDEL\` sweep keyed off consumer-group ack state) and confirm it's safe given \`delivery_row_id\` is the only durable reference back to \`webhook_deliveries\` (i.e. trimming a stream entry doesn't lose anything permanently -- the row itself, and its \`status\`, are the source of truth; the stream entry is just the queue signal).

nazarli-shabnam:
## What
The \`webhook_events\` Redis Stream (written by both \`routers/webhooks.py\`'s ingestion path and, as of #409, \`webhook_requeue_sweep.py\`'s retry path) had no retention/trimming policy — entries accumulate forever unless read+acked, and nothing ever \`XTRIM\`s it. CodeRabbit flagged this on PR #338 (the stream's original introduction) and asked for a defined recovery policy before any trimming is added.

## Design decision
Went with **\`XADD ... MAXLEN ~ N\`** (approximate trimming built into the XADD call itself) over a separate periodic \`XTRIM\`/\`XDEL\` sweep loop — no new asyncio loop, no new config key, just one kwarg on the two existing \`xadd\` calls. \`N = 50,000\`, sized as generous headroom over normal throughput; hardcoded (not DB-configurable), matching this file's existing \`_MAX_AGE_HOURS\`/\`_BATCH_LIMIT\` constants rather than adding Settings-page wiring for a value that shouldn't need runtime tuning.

**The defined recovery policy CodeRabbit's original concern demanded**: trimming a stream entry never loses anything on its own — \`webhook_deliveries\`'s row + \`status\` column is the actual source of truth, the stream entry is just the queue signal — but until now nothing would catch a row whose entry got trimmed before the consumer group ever read it (it would sit at \`status='queued'\` forever with no automatic recovery). So I extended \`webhook_requeue_sweep.py\` (issue #409's existing sweep) to also re-\`XADD\` a row still \`queued\` more than 30 minutes after receipt — normal \`XREADGROUP\` pickup latency is seconds, so anything still queued that long either had its entry trimmed or never actually landed. Re-adding it is safe: the consumer only flips a row to \`processed\` after it's actually applied, so a still-\`queued\` row is guaranteed not yet consumed.

## Change
- \`routers/webhooks.py\`: \`_WEBHOOK_STREAM_MAXLEN = 50_000\`, added to its \`xadd\` call.
- \`webhook_requeue_sweep.py\`: matching \`_STREAM_MAXLEN\` constant + \`maxlen\`/\`approximate\` on its own \`xadd\`; widened its query to catch \`status='queued'\` rows older than a new \`_STUCK_QUEUED_MINUTES = 30\` threshold, not just \`'queue_failed'\` ones.
- Updated one existing test's exact-call assertion for the new kwargs; added a test proving a stuck-queued row gets re-enqueued (and a matching "not too soon" case already existed and still passes).

## Verification
\`pytest apps/api apps/worker\` — full suite passes (1216 passed, 3 skipped).

Closes #440

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved webhook delivery reliability by automatically retrying deliveries that remain queued for an extended period.
  * Prevented delayed webhook processing from causing unbounded event backlog growth.
  * Preserved delivery status tracking while recovering eligible webhook events for processing.


nazarli-shabnam:
### Code review

Found 1 issue:

1. The new stuck-queued re-XADD path doesn't account for `webhook_deliveries` rows that `event_consumer.py` leaves at `status='queued'` **permanently by design**, not transiently. Two branches in `_process_entry` `xack` the stream entry but deliberately never flip status to `'processed'`: an unresolved `tenant_id` (https://github.com/nazarli-shabnam/clevis/blob/07874e1e520c772f10032fc37720745b6039ce04/apps/worker/src/event_consumer.py#L338-L344), and `membership`/`team` events with no normalizer yet (https://github.com/nazarli-shabnam/clevis/blob/07874e1e520c772f10032fc37720745b6039ce04/apps/worker/src/event_consumer.py#L346-L352). The new sweep query has no event_type/consumer-state filter to exclude these:

https://github.com/nazarli-shabnam/clevis/blob/07874e1e520c772f10032fc37720745b6039ce04/apps/api/src/services/webhook_requeue_sweep.py#L53-L54

So once one of these rows crosses `_STUCK_QUEUED_MINUTES` (30 min), the sweep re-XADDs it, the consumer immediately acks it again without changing status, and the next sweep tick (every `gap_heal`-style interval this runs on) re-XADDs it again — forever. Not data loss, but an infinite, unbounded retry loop that keeps writing to the very stream this PR is trying to bound with `MAXLEN`, for every row that ever lands with a null tenant (a real, ongoing occurrence per the test suite's own `_make_row(db, tenant_id=None)` default) or a stale membership/team backlog row. Worth excluding rows event_consumer.py leaves queued on purpose (e.g. by event_type, or by tracking a distinct 'ack'd but not normalized' status) before this ships.

nazarli-shabnam:
Fixed in 6628da7: the sweep's stuck-queued query now excludes rows event_consumer.py deliberately leaves 'queued' forever (`tenant_id IS NULL`, or `event_type` in the not-yet-normalized set), so it can't re-XADD the same row every tick indefinitely. Added two tests covering both exclusions. Full suite (1218 tests) passes.

